PR #11, opened 2019-11-19
Added Link Array

Alonso-JAMM:
A link array now can be created from within Assembly 4.

In order to create an array, a App::Link object needs to be selected first. Then a name for the new array can be given in the pop-out dialog. 

Once the array is created, its parameters (like size and type). The wiki contains more information information about link array https://www.freecadweb.org/wiki/Draft_LinkArray


Zolko-123:
Hello, this is great news. Before merging I'll download and test it locally.

PR #23, opened 2020-01-16
Fix broken link to Assembly4 Instructions.md

justinclift:


Zolko-123:
please use relative links (and not absolute http:// ones)

justinclift:
No worries.  Saw the absolute one for the .stp file just below it, and mistakenly copied that approach.  Will remember for the next time. :smile:

PR #55, opened 2020-03-24
A4: update link array from Draft

vocx-fc:
In a recent version of Draft v0.19, the link array creation function has the argument `use_link`, instead of `useLink` to comply with Python style guidelines (PEP8).

Besides that, Draft now has more useful array commands, that could be used, for example, Draft_OrthoArray, Draft_PolarArray, and Draft_CircularArray. They can create App::Links as well.


Zolko-123:
if this PR is still valid, can you please make it against the -development branch and not against -master ?

vocx-fc:
This is still valid, otherwise I think the Array tool inside Assembly4 is still broken.

New pull request #58.

PR #56, opened 2020-03-24
A4: subclass the new Draft_OrthoArray tool

vocx-fc:
This showcases how to add the new [Draft_OrthoArray](https://wiki.freecadweb.org/Draft_OrthoArray) tool from the Draft Workbench into Assembly4.

It imports the required modules, and then subclasses the required class, and adds a bit of specific code to only be able to select `App::Link` objects.

The original `Draft_OrthoArray` launches its own task panel interface in order to change the properties of the created array. It is expected that in the future we will be able to show a bit more advanced behavior, that is, previews of the array.

What is missing is specific code that is able to rename the created array, and place it in the `Model` object of Assembly4. To do this, probably `Draft_OrthoArray` needs to be modified to accept a final callback from any subclass.

A very similar process as the one described in this commit could be done to include the `Draft_PolarArray` and `Draft_CircularArray` tools in Assembly4.

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -131,7 +131,8 @@ class Assembly4Workbench(Workbench):
                                 "Asm4_importDatum", 
                                 "Asm4_placeDatum", 
                                 "Asm4_updateAssembly",
-                                "Asm4_newLinkArray"]
+                                "Asm4_newLinkArray",
+                                "Asm4_OrthoArray"]
         # commands to appear in the Assembly4 menu 'Assembly'
         self.itemsMenu =      [ "Asm4_newModel",   
                                 "Asm4_newBody", 
@@ -152,6 +153,7 @@ class Assembly4Workbench(Workbench):
                                 "Asm4_placeDatum", 
                                 "Asm4_importDatum", 
                                 "Asm4_newLinkArray",
+                                "Asm4_OrthoArray",
                                 "Asm4_addVariable", 
                                 "Asm4_Animate", 
                                 "Asm4_updateAssembly"]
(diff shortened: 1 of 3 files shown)


Zolko-123:
Hi, I was wondering whether you had forgotten about Asm4.Array, glad to see that you're still on it.

# Only active when a App::Link is selected

I think that check-part also accepts fasteners, can you please verify and update the comment ?

Zolko-123:
also:

```
    def IsActive(self):
        if App.ActiveDocument:
            # Only active when a App::Link is selected
            selectObj = self.checkPart()
            if selectObj:
                return True
        else:
            return False
```
if there is an active document but check-selection returns false, what is the result ? I think the last _else_ should be removed: either we have a path to return True, and if not we fall-back to the end where we return False.

vocx-fc:
> I think the last else should be removed:

I guess you are right, although I didn't worry too much about the code. I basically just took what you had and quickly changed it to make it work. I'm not sure if an explicit `False` is needed, but whenever I implement `IsActive`, I normally add it.

I guess it would be simpler like this
```py
 def IsActive(self):
        if App.ActiveDocument and self.checkPart():
            return True
        else:
            return False
```

Zolko-123:
If this PR is still valid, can you please make it against the -development branch, and not against -master ?

vocx-fc:
This pull request does not strictly need to be merged because it doesn't solve a bug. It is just for demonstration purposes. It shows how the new Draft array tools can be added to Assembly4.

The original tool is called `Asm4_newLinkArray`, and this pull request introduces `Asm4_OrthoArray`. In the future, the latter could replace the former.

New pull request #59.

PR #58, opened 2020-04-18
A4: update link array from Draft

vocx-fc:
In a recent version of Draft v0.19, the link array creation function has the argument `use_link`, instead of `useLink` to comply with Python style guidelines (PEP8).

Besides that, Draft now has more useful array commands, that could be used, for example, Draft_OrthoArray, Draft_PolarArray, and Draft_CircularArray. They can create App::Links as well.


Zolko-123:
Hi vocx, is this still a valid PR ? Do you think you could propose something better suited using the new Draft_OrthoArray & friends ?

vocx-fc:
Did you try the code? This is a very small change, that just fixes what broke around February 2020, so it works as before. The following pull request #59, that one introduces the new arrays.

PR #59, opened 2020-04-18
A4: subclass the new Draft_OrthoArray tool 

vocx-fc:
This showcases how to add the new [Draft_OrthoArray](https://wiki.freecadweb.org/Draft_OrthoArray) tool from the Draft Workbench into Assembly4.

It imports the required modules, and then subclasses the required class, and adds a bit of specific code to only be able to select `App::Link` objects.

The original `Draft_OrthoArray` launches its own task panel interface in order to change the properties of the created array. It is expected that in the future we will be able to show a bit more advanced behavior, that is, previews of the array.

What is missing is specific code that is able to rename the created array, and place it in the `Model` object of Assembly4. To do this, probably `Draft_OrthoArray` needs to be modified to accept a final callback from any subclass.

A very similar process as the one described in this commit could be done to include the `Draft_PolarArray` and `Draft_CircularArray` tools in Assembly4.

---

This pull request does not strictly need to be merged because it doesn't solve a bug. It is just for demonstration purposes. It shows how the new Draft array tools can be added to Assembly4.

The original tool is called `Asm4_newLinkArray`, and this pull request introduces `Asm4_OrthoArray`, `Asm4_PolarArray`, and `Asm4_CircularArray`. In the future, these tools could replace the original one.

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -231,6 +231,7 @@ class Assembly4Workbench(Workbench):
                                 "Asm4_importDatum", 
                                 "Asm4_newLinkArray",
                                 "Asm4_OrthoArray",
+                                "Asm4_PolarArray",
                                 "Asm4_addVariable", 
                                 "Asm4_Animate", 
                                 "Asm4_updateAssembly"]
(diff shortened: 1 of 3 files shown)


vocx-fc:
I haven't updated this pull request because there was no comment on it. Now there seems to be a conflict.

The code should work, the only issue is that the array will not be created inside the `Model` object. To me it's not a big deal, but if you want to do that, then a bit of additional code needs to be added to move the new object inside of it.

PR #112, opened 2020-09-09
Fix highlight of insert link button in IsActive function + Adding filtering feature to insert link dialog

abetis:
The Asm4.checkModel() was called without (), so the button was always active


abetis:
Apparently I've added 2 changes to the same PR...

Zolko-123:
Could you please give some details what sort of filtering this does ? Is it a text search on the name ?

I might have made a mistake about the __pycache__ in the gitignore, dunno, should not be too important.

abetis:
There was no .gitignore file in the directory, so I just added one to branch, so the cache will not appear in the changed files list.

The filtering is done based on the text in the widgets added to the partList.
Tee filter only hides/shows the relevant entries, it doesn't change the list.

Zolko-123:
This is very useful, thank-you

abetis:
I have some ideas to re-do the UI so the connection points could be selected from the 3D view, such as in T-Flex, but for now the simple filter might help those who have many parts to select (me)...

PR #113, opened 2020-09-10
Override 3D view click to select linked part entry in tree view

abetis:
Override 3D view click to select linked part entry in tree view
instead of selecting the body of the part.
Improves the efficiency of work, the assembly properties are faster accessible.

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -105,6 +105,7 @@ class Assembly4Workbench(Workbench):
         import Asm4_Measure        # Measure tool in the Task panel
         import makeBomCmd          # creates the parts list
         import HelpCmd             # shows a basic help window
+        import treeSelectionOverride    # ovveride 3D view click to select assembly entry in the tree view
         
         # create the toolbars and menus, nearly empty, to decide about their position
         self.appendToolbar("Assembly",["Asm4_newModel"])
--- a/InitGui.py
+++ b/InitGui.py
@@ -31,7 +31,7 @@ asm4wb_icons_path = os.path.join( asm4wbPath, 'Resources/icons')
 global main_Assembly4WB_Icon
 main_Assembly4WB_Icon = os.path.join( asm4wb_icons_path , 'Assembly4.svg' )
 
-
+import treeSelectionOverride as selectionOverride
 
 """
     +-----------------------------------------------+
@@ -64,6 +64,7 @@ class dropDownCmdGroup:
 class Assembly4Workbench(Workbench):
 
     global main_Assembly4WB_Icon
+    global selectionOverride
     MenuText = "Assembly 4"
     ToolTip = "Assembly 4 workbench"
     Icon = main_Assembly4WB_Icon
@@ -74,10 +75,12 @@ class Assembly4Workbench(Workbench):
 
     def Activated(self):
         "This function is executed when the workbench is activated"
+        selectionOverride.Activate()
         return
 
     def Deactivated(self):
         "This function is executed when the workbench is deactivated"
+        selectionOverride.Deactivate()
         return 
 
     def GetClassName(self): 
@@ -105,7 +108,6 @@ class Assembly4Workbench(Workbench):
         import Asm4_Measure        # Measure tool in the Task panel
         import makeBomCmd          # creates the parts list
         import HelpCmd             # shows a basic help window
-        import treeSelectionOverride    # ovveride 3D view click to select assembly entry in the tree view
         
         # create the toolbars and menus, nearly empty, to decide about their position
         self.appendToolbar("Assembly",["Asm4_newModel"])
--- a/libAsm4.py
+++ b/libAsm4.py
@@ -113,6 +113,13 @@ def checkModel():
             retval = model
     return retval
 
+def isLinkToPart(obj):
+    if obj.TypeId == 'App::Link' and hasattr(obj.LinkedObject,'isDerivedFrom'):
+        if  obj.LinkedObject.isDerivedFrom('App::Part') or obj.LinkedObject.isDerivedFrom('PartDesign::Body'):
+            return True
+    else:
+        return False
+
 
 # get from the selected datum the corresponding link
 def getLinkAndDatum():
@@ -125,10 +132,9 @@ def getLinkAndDatum():
         for objStr in parentAssembly.getSubObjects():
             # the string ends with a . that must be removed
             obj = App.ActiveDocument.getObject( objStr[0:-1] )
-            if obj.TypeId == 'App::Link' and hasattr(obj.LinkedObject,'isDerivedFrom'):
-                if  obj.LinkedObject.isDerivedFrom('App::Part') or obj.LinkedObject.isDerivedFrom('PartDesign::Body'):
-                    # add it to our tree table if it's a link to an App::Part ...
-                    childrenTable.append( obj )
+            if isLinkToPart(obj):
+                # add it to our tree table if it's a link to an App::Part ...
+                childrenTable.append( obj )
 
         selObj = Gui.Selection.getSelection()[0]
         # a datum is selected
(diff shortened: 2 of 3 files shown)


Zolko-123:
Hi,

yes, adding a selection observer is the good way to do it. But shouldn't this be in the libAsm4 file ? Did you try to put _asm4SelObserver()_ into libAsm4 as _selObserver()_ , to be invoked as  _Asm4.selObserver()_

Actually, when you look at the Asm4_Measure tool, there is a selection observer, but it's probably not done in a very good way.

Zolko-123:
What I mean is: you invoke the selection observer from _InitGui_, does that mean that treeSelectionOverride will always be active ? Aren't there any unintended consequences ? Should a selection observer be removed once it is not used ? 

If this only concerns (for now) the placement of linked parts, shouldn't this rather be invoked from _placeLinkCmd.py_ ? Did you think about making the selection observer in placeLinkCmd.py, and create it when calling the task, then removing it when quitting the task ?

abetis:
You're right about removing the observer. I forgot that there are other workbenches out there...
Should remove the observer when Assembly4 is deactivated, I will send an update soon.

While you're in Assembly workbench, I would always want to select the linked part instead of the body. Don't you?
That's the more common use case. If someone needs to edit the body of the part, he could click on the [+] in the tree and get to the body.

placeLinkCmd handles the placement/insert UI and related functionality, I don't see a reason to unite those 2 different functions (placement and 3D view click override)

Zolko-123:
> I would always want to select the linked part instead of the body

hold-on: there is a special case for creating the Hole-Axis datum, where I explicitly want to select an edge. There is also the case when several LCS overlap: how will you make the difference ? 

So I'd suggest for now to make this function only for the placement of a link, and remove the observer when the placement task is finished. We will probably have to fine-tune this functionality.

abetis:
You are right about the creation of axis datum, didn't think about it.

The use case where I wanted that feature is not only for placing a link.
I mostly assemble imported STEPs with lots of holes and connection points. Here's an example:
https://www.gobilda.com/1120-series-u-channel-14-hole-360mm-length/

Since I can't add LCS to all the holes in the part, I generally add few hole LCSs on different sides, or sometimes change the placement of the LCS_Origin of the imported STEP to sit on one of the holes or align on the side of the part.
After I add that part to assembly with proper rotation, the first thing I have to do is go to the assembly properties and play with the attachment position to place the part in proper offset.
Sometime I see collision between the parts and move some of them around by changing the attachment position (there are lots of holes where I can connect them).
With all that process I click alot on the 3D view to select the parts since the tree view pretty large and it's not practical to scroll all that, especially when the tree entries unfold on any click on the part/body in the 3D view.

I think those who play with the animation will also like that working mode.
Actually I think that mode is not needed when preparing the parts for assembly (adding LCSs), but it is needed when dealing with assembly/animation only.

2 options I can think off:
1. Use keyboard modifiers to temporary disable that feature. Ctrl+click will be able to select faces/edges.
2. Have that option with switchable toolbar, allowing the users to enable and disable that mode.

We can start with option 2, having it disabled by default since that's the current behavior and after some time ask people what mode do they mostly use, maybe we will set it enabled by default if most of the people prefer to use that mode.
Or one day we might have global FreeCAD configuration for Assembly4 that will have that setting.

What do you say?

The functionality is already implemented so 80% is done. enabling/disabling it - it's the last 20%.

Zolko-123:
Hum ... I'm not sure to understand how this works: ideally, I thought that when placing  a part, and selecting an LCS in the 3D window, it would recognize the LCS and the link to which it belongs, and pre-select them from the presented lists. But this doesn't seem to be the case. 

Could you please explain better how to use this modification.

abetis:
Pre-selection works as it was before and that code doesn't touch it.
The code listens to the addSelection (when something was selected in the tree or clicked on the 3D view).
The addSelection gets 3 parameters: document name, object name and subObject name.
According to realthunder, the getSubObjectList returns a list of objects connecting the selected object to the document.
The code goes from the bottom of the list (LCS entry first) up to the top, looking for the link entry (the isLinkToPart function).
When found it, clear the selection and select the link entry instead.

abetis:
Actually I see there is a bug...
I have a model with sub-sub-links and those doesn't work since the code looks for the inner objects in the top document instead of the linked sub-document.
Will check if the getSubObjectList returns correct list in that case or maybe I have to add more checks if there is a link on the way in the list

Zolko-123:
This forbids the use of the datum HoleAxis, which I find very useful. Therefore, this can't be activated at the Asm4 WB level, and I'll place this into _placeLinkCmd_.

abetis:
What do you mean forbids?
What is the flow? Selecting a hole to add the hole axis?

I see that in documents that I have only model and body under the model (imported STEP), click on the body selects the body in the tree, which is very useful.

Please send an example of the file.

Zolko-123:
With this selectionObserver activated, one cannot select an edge, the entire part is always selected. I moved it to `placeLinkCmd()`

abetis:
I still don't understand the flow...
When do you have to select the edge?
To add a axis?

As I wrote before, I think the feature should have a button to enable/disable it.
Do you want me to add such toolbar button?

Zolko-123:
You may want to highlight an edge for whatever reason (design ...) but especially when you want to add a datum axis on a circular edge to place fasteners: _New Hole Axis_. I've now adapted your selection observer such that when you select a link in the window, it selects it also in the drop-down and presenting the available datums. 

Please update the development branch to test this functionality. I think it follows your desire but is not intrusive for other use-cases

abetis:
Can you send me a test file that you use?
The files I use, that include one part, which was imported from STEP, I can select edges, faces, points and whatever you need to add a datum.
Those files don't have links, so the code doesn't select the link object.
For my usage I don't see any problem...

Zolko-123:
here : [asm_tuto2.zip](https://github.com/Zolko-123/FreeCAD_Assembly4/files/5279105/asm_tuto2.zip)

Zolko-123:
yes, actually may-be a general behaviour with a global toggle could also be nice.

abetis:
I still don't understand what do you try to achieve...
Here's how I see it:
Using the asm_tuto2 you sent.
I want to change the arm001 offset in the assembly.
For any reason, want to have it a bit higher, add a spacer, check how high I can get it up in order to change the base001 pad where it's connected...
I click on the top of the arm1.
Where do I get in the tree?
arm001 --> Body002 --> Chamfer
I don't need it... I'm assembling it now, not changing the arm001 part...
Now I have to select the arm001 in the tree in order to get to the assembly properties.
Find the "Attachment Offset --> Position --> z" property, change it to 15.

With the override features anywhere I click on arm001 I stay on the link level in the tree.
From there I can go directly to the "Attachment Offset" and other assembly related properties and actions, such as Move/Attach part in the assemby.
As I see it, if you're in the assembly, stick to the assembly behavior. Don't want to see the part creation operations such as chamfer/pad and so on, those are part of the part creation process.

I understand that if you want to change the parts, which are in the same document, than that feature gets in the way and you can't select faces to add datums and other stuff.
All the assemblies I have are links to external parts, so if I want to change the part, I have to switch to the part file, there are no links in that file, so everything works fine without the need to disable the feature.
The problem exist only for files that have both parts and the assembly in the same file, like in the asm_tuto2.

I've tries to follow the code in the development branch and didn't understand what you did with the observer there.
I've switched to your development branch and I don't see the override feature working.
Should I enable something in python to make it work?

I will add a toolbar button to my stream to enable/disable it.
This way everybody could switch to whatever mode they want.

Zolko-123:
> _I've switched to your development branch and I don't see the override feature working. Should I enable something in python to make it work?_

did you try with the `placeLinkCmd` command ? I have moved it to there.

As for why I changed your first proposed behaviour, try to create an LCS and edit its attachment in the assembly. For example by making a master sketch first. With Assembly4, it's true that we're doing assemblies, but also the design and placement of sketches and datum objects, and the global selection is in the way there.

> _I will add a toolbar button to my stream to enable/disable it. This way everybody could switch to whatever mode they want._

yes, that would be the ideal behaviour

PR #114, opened 2020-09-12
Show/Hide LCS in objects and child objects.

abetis:
Please see the code.
Icons are still missing.
Currently accessible from context menu (right-click) when a part is selected on assembly level or the whole "Model" entry in the tree.

It's slow when there are many objects, I don't like the API we have today with the getSubObjects and getSubObject(name) functions that slow down everything related to tree entries, including the pre-selection I think.
Maybe will look on the code later when I have time...

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -105,6 +105,7 @@ class Assembly4Workbench(Workbench):
         import Asm4_Measure        # Measure tool in the Task panel
         import makeBomCmd          # creates the parts list
         import HelpCmd             # shows a basic help window
+        import showHideLcsCmd      # shows/hides all the LCSs
         
         # create the toolbars and menus, nearly empty, to decide about their position
         self.appendToolbar("Assembly",["Asm4_newModel"])
@@ -238,7 +239,10 @@ class Assembly4Workbench(Workbench):
                                 "Asm4_importDatum"   ,
                                 "Asm4_placeDatum"    ,
                                 'Asm4_FSparameters'  ,
-                                'Asm4_placeFastener' ] 
+                                'Asm4_placeFastener' ,
+                                'Separator'          ,
+                                'Asm4_showLcs'       ,
+                                'Asm4_hideLcs']
         # commands to appear in the 'Create' sub-menu in the contextual menu (right-click)
         self.itemsCreateMenu = ["Asm4_newSketch",  
                                 "Asm4_newBody", 
--- a/libAsm4.py
+++ b/libAsm4.py
@@ -39,6 +39,8 @@ partInfo =[     'Description',                  \
                 'SupplierDescription',          \
                 'SupplierReference' ]
 
+containerTypes = [  'App::Part', \
+                    'PartDesign::Body' ]
 
 """
     +-----------------------------------------------+
--- a/libAsm4.py
+++ b/libAsm4.py
@@ -482,3 +482,30 @@ def splitExpressionDatum( expr ):
 
 
 
+"""
+    +-----------------------------------------------+
+    |        Selection Helper functions             |
+    +-----------------------------------------------+
+"""
+def getModelSelected():
+    if App.ActiveDocument.getObject('Model') and App.ActiveDocument.Model.TypeId == 'App::Part':
+        selection = Gui.Selection.getSelection()
+        if len(selection)==1:
+            selObj = selection[0]
+            if selObj.Name == 'Model' and selObj.TypeId == 'App::Part':
+                return selObj
+    return False
+    
+def getSelection():
+    # check that there is an App::Part called 'Model'
+    if App.ActiveDocument.getObject('Model') and App.ActiveDocument.Model.TypeId == 'App::Part':
+        selection = Gui.Selection.getSelection()
+        if len(selection)==1:
+            selObj = selection[0]
+            # it's an App::Link
+            if selObj.isDerivedFrom('App::Link') and selObj.LinkedObject.TypeId in linkedObjTypes:
+                return selObj
+    return None
+
+# type of App::Link target objects we're dealing with
+linkedObjTypes = [ 'App::Part', 'PartDesign::Body' ]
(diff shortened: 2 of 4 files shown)


Zolko-123:
yeah, that's really good, thank-you. Would you mind if I put that in a separate menu (Assembly > shwo/hide LCS) entry instead of a contextual menu command ?

Zolko-123:
the function _getModelSelected()_ seems to do exactly what the function _checkModel()_ already does. Can you confirm ?

abetis:
No, those functions are different. Similar naming might be confusing...
The `checkModel` returns the model object when it's available in the document - to see if Assembly4 structure exist
The `getModelSelected` returns the model object when it's selected in the tree view.

Zolko-123:
`checkModel()` returns the Asm4 Model if it's available, and `getModelSelected()` returns the Asm4 Model if it's available **and** selected: is there a very strong reason for this alternate functionality ? If anything, `getPartSelected()` would be a better alternative I think.

Also:
`linkedObjTypes = [ 'App::Part', 'PartDesign::Body' ]`
`containerTypes = [  'App::Part', 'PartDesign::Body' ]`
It would be good, for maintenance, to not duplicate functionality.

Zolko-123:
Actually, why do we need to check whether an Asm4 Model is present ? Can't we just show and hide the LCS for a selected App::Part (and/or PartDesign::Body) in any case ? 

What do you think of the following: create a new function `getSelectedContainer()` that returns the App::Part or PD:Body container if one is selected, and show/hide all its LCS ? And if nothing is selected, show/hide all LCS if an Asm4 Model is present. Else, do nothing.

abetis:
> `checkModel()` returns the Asm4 Model if it's available, and `getModelSelected()` returns the Asm4 Model if it's available **and** selected: is there a very strong reason for this alternate functionality ? If anything, `getPartSelected()` would be a better alternative I think.
I probably don't see the whole picture since those functions are on parallel streams, I don't see the merged code right now.
I've added the `getModelSelected()` to know if Model tree entry was selected since it needs a special processing to dive into all sub-objects to show/hide the LCSs.
The `checkModel()` function was in the code as far as I remember, or maybe I've moved the code you used to a function since I needed the same functionality, you use that function to check if Asm4 model is in the document to enable/disable the toolbar buttons...

> Also:
> `linkedObjTypes = [ 'App::Part', 'PartDesign::Body' ]`
> `containerTypes = [ 'App::Part', 'PartDesign::Body' ]`
> It would be good, for maintenance, to not duplicate functionality.
I don't see that in my code. Maybe a product of a merge again?
Ofcause if there is such duplication, only one set should be left

Zolko-123:
`linkedObjTypes = [ 'App::Part', 'PartDesign::Body' ]`

it's at the end of libAsm4

abetis:
> Actually, why do we need to check whether an Asm4 Model is present ? Can't we just show and hide the LCS for a selected App::Part (and/or PartDesign::Body) in any case ?

What is the top-level container of all the linked objects?
I've added that special case since something didn't work the same way as with a selected object. Maybe I should revisit the code.

> What do you think of the following: create a new function `getSelectedContainer()` that returns the App::Part or PD:Body container if one is selected, and show/hide all its LCS ? And if nothing is selected, show/hide all LCS if an Asm4 Model is present. Else, do nothing.

What if I right-click on Variables or Origin entry? I don't think we should allow show/hide LCSs for that case.
What if we will end the toolbar buttons at the end and the user don't have any selection at all, should we do something?
I don't want to guess. If you selected a part and clicked hide-lcs, so we should hide the LCS for that part, if you select the top-level container, be it the Model or something else, than we should hide-lcs for all the parts in that container.

Do you have plans to remove the "Model" entry?
You add it when creating the Assembly4 model, so that's a good anchor to know if we're in Assembly4 document or somewhere else.

abetis:
> `linkedObjTypes = [ 'App::Part', 'PartDesign::Body' ]`
> 
> it's at the end of libAsm4

I've added the containerTypes, based on git history you added the linkedObjTypes in the "UI in task panel" commit.
We probably had the same intention in parallel changes. Only one should be left of cause.

Zolko-123:
> What if I right-click on Variables or Origin entry?

I think that : either a valid container is selected, and then we apply show/hide on all LCS inside and inside its children, or if nothing is selected but an Asm4 Model is present, then we show/hide all LCS in the Asm4 Model, else we do nothing. So if a Variables of Origin is selected, we do nothing

> Do you have plans to remove the "Model" entry? You add it when creating the Assembly4 model, so that's a good anchor to know if we're in Assembly4 document or somewhere else.

you're right, and that was the purpose. Somewhat hard-coded but quite practical, therefore I don't see why change that. We need that to insert and place linked parts (assembly !), but we don't actually need it to insert a new sketch or a datum: in these cases, we only check whether a container is selected and create the object inside that container, even if no Asm4 Model is present.

The idea is: only check for an Asm4 Model if **absolutely** necessary.

abetis:
So I should not use the Asm4.getSelection() function?
The Asm4.getSelection() checks if the model is present.
If nothing is selected / or Asm4.getSelection() return None, I should still check if the model is present and find it to get to the children.
Need the model to touch the LCSs there. The document might include other objects that might have lines and points that should not be changed.

Maybe I don't fully understand you...

Zolko-123:
yes, you're right, `Asm4.getSelection()` is the good choice: since the point is to show/hide LCS in linked objects, we're talking about an assembly, therefore checking whether there is an Asm4 Model first is good. Sorry for the confusion.

abetis:
@Zolko-123 
Why did you change those lines, why is it better?
-        for linkObj in obj.LinkedObject.Document.Objects:
+        for objName in obj.LinkedObject.getSubObjects():
+            linkObj = obj.LinkedObject.Document.getObject(objName[0:-1])

Zolko-123:
ha haaa !!!! Mistery ....

it's because your line returns all objects in the document of the linked object, while mine returns only the objects contained in the object itself. If each part is in its own document, the result is the same, but when a document contains several parts (as in an assembly in a single document) then all datums of that document are shown/hidden, not only the ones from the selected part.

PR #117, opened 2020-09-17
Configuration engine

abetis:
Hi. This is till work in progress, but the engine works for save and restore (right click context menu).
Checked with different levels of sub-sub-assembly.
Please take a look on the code.

Need the icons again.

Need to add UI to select configuration to restore and get a configuration name from the user to save it.
Will explore how to do it on the tasks panel. Will add the rest of the code when I understand how to do it.

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -109,7 +109,8 @@ class Assembly4Workbench(Workbench):
         import makeBomCmd          # creates the parts list
         import HelpCmd             # shows a basic help window
         import showHideLcsCmd      # shows/hides all the LCSs
-        
+        import configurationEngine  # save/restore configuration
+
         # create the toolbars and menus, nearly empty, to decide about their position
         self.appendToolbar("Assembly",["Asm4_newModel"])
         # self.appendToolbar("Design",["Asm4_newPart"])
@@ -245,7 +246,10 @@ class Assembly4Workbench(Workbench):
                                 'Asm4_placeFastener' ,
                                 'Separator'          ,
                                 'Asm4_showLcs'       ,
-                                'Asm4_hideLcs']
+                                'Asm4_hideLcs'       ,
+                                'Separator'          ,
+                                'Asm4_saveConfiguration',
+                                'Asm4_restoreConfiguration']
         # commands to appear in the 'Create' sub-menu in the contextual menu (right-click)
         self.itemsCreateMenu = ["Asm4_newSketch",  
                                 "Asm4_newBody", 
--- a/configurationEngine.py
+++ b/configurationEngine.py
@@ -1,0 +1,85 @@
+#!/usr/bin/env python3
+# coding: utf-8
+#
+# configurationEngine.py
+#
+# The code to save and restore configurations, using spreadsheets
+
+import FreeCADGui as Gui
+import FreeCAD as App
+import libAsm4 as Asm4
+
+HEADER_CELL             = 'A1'
+NAME_CELL               = 'A2'
+OBJECTS_START_ROW       = '5'
+OBJECT_NAME_COL         = 'A'
+ATTACHMENT_POS_X_COL    = 'B'
+ATTACHMENT_POS_Y_COL    = 'C'
+ATTACHMENT_POS_Z_COL    = 'D'
+ATTACHMENT_AXIS_X_COL   = 'E'
+ATTACHMENT_AXIS_Y_COL   = 'F'
+ATTACHMENT_AXIS_Z_COL   = 'G'
+ATTACHMENT_ANGLE_COL    = 'H'
+
+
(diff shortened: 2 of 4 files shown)


abetis:
Feature is ready, besides the icons...
There is a limitation that if you save the whole "Model", you can't restore individual objects from that configuration, or any other combinations.
You have to restore exactly the same object/link that you saved.
Nothing will crash, just the configuration will not be found.

I've asked realthunder if there is an API that will return full object name from the top level to the sub-sub link position.
If there is such API, than the feature could use that full naming and be able to find individual objects even if the whole model is saved.

Even with those limitations the feature is very usable.
I have some ideas how to use it for explode feature that I will check next...

abetis:
Thanks for realthunder, found the solution that allows saving the whole Model and restore a specific object configuration.
The feature is complete as far as I see it.

Zolko-123:
This looks really cool. 

But can you please fix the _getModelSelected()_ function ? It seems to me that it returns the Asm4 Model like _checkModel()_ but only if it's actually selected: I don't see the use-case for this extra work.

abetis:
First time I resolve a conflict on GitHub, hope it went well...

That's exactly what the function does, checks if the "Model" entry was selected, if true, return the model, if not, return None.
The Show/Hide LCSs feature and Configuration Engine both needs to know when the selection is the Model entry or anything else. There is a different processing for the cases...

Zolko-123:
I'm testing this now (didn't have time before). This is really nice, seems to work fine, thank-you very much. I have a couple of comments :

1) could we add to the configuration the visibility of a part ?
`obj.ViewObject.Visibility = True or False`
This would be really handy to make user manuals: we build the entire model, and then make configurations for each build step, hiding the parts that are not assembled at the considered time.

2) Why is the description of a configuration mandatory ? This should not be, but it crashes when left empty, or when only a number is typed there

3) saving the rotation as axis X, Y, Z and angle might not be ideal. When asking FreeCAD about the AttachmentOffset we get:
`obj.AttachmentOffset = Placement [Pos=(27,0,0), Yaw-Pitch-Roll=(0,0,0)]`
this is: 
`Yaw = obj.AttachmentOffset.Rotation.toEuler()[0] ... 1, 2`
and restore it with
`rotation = App.Rotation(Yaw, Pitch, Roll)`
and change the column headers accordingly. What do you think ? Like that, it should even be possible to edit the configuration spreadsheet manually

Zolko-123:
Let's call the colmuns:


```
    doc.set( OBJECT_NAME_COL +       headerRow, 'ObjectName' )
    doc.set( OBJECT_VISIBILITY_COL + headerRow, 'Visibility' )
    doc.set( OFFSET_POS_X_COL +      headerRow, 'Base_X' )
    doc.set( OFFSET_POS_Y_COL +      headerRow, 'Base_Y' )
    doc.set( OFFSET_POS_Z_COL +      headerRow, 'Base_Z ')
    doc.set( OFFSET_YAW_COL +        headerRow, 'Rotation_Yaw' )
    doc.set( OFFSET_PITCH_COL +      headerRow, 'Rotation_Pitch' )
    doc.set( OFFSET_ROLL_COL +       headerRow, 'Rotation_Roll' )
```

What do you think ?

abetis:
Hi.
I don't have time for that project either during that crazy period.

The code is not in front of me, but as far as I remember, I tried to use the angle in the same manner as it appears in the properties. The angle there is definitely not yaw/pitch/roll, but one float value as "angle" and 0.0 - 1.0 value for every axis.
If there is an easy way to convert logical representation of Yaw/Pitch/Roll to that encoding, so definitely is should go that way.
I remember checking how to do it, but couldn't find the needed code or property.

Adding visibility is a good idea.

If you have the time, feel free to change it, I will catch it up later when I'll have the time to get back to that project. Unfortunately don't know yet when it will happen.

Zolko-123:
OK, thanx for the answer. I'll do the changes

abetis:
Forgot to answer about the description of the configuration...
I thought that when someone will have too many configurations, it will not be easy to determine which is what.
The "file name" that can be used to give short description can't be too long if you want to save some notes.
If you feel that its redundancy, you can remove that restriction

Zolko-123:
Actually, you can't leave the description empty because then the cell is not set and when you try to read it it throws an error. So when the user leaves the description empty we now cope the name into the description. Works, but is visually sub-optimal.

BTW, the code works very well, thank-you.

abetis:
Try writing an empty string instead "", or a space " ".
On read it fails probably because the cell doesn't exist.
Writing a space will create the cell, but on read it should probably be trimmed before checking if there is a description that is added after the file name in the restore configuration selection, so it will not appear as "filename ( )".

PR #124, opened 2020-09-25
Enable/Disable 3D view Link selection mode

abetis:
Adding toolbar button to enable/disable select mode.
Unfortunately it looks like there is no way to update the icon after addition of the command on initialization.
All I could do is add a printout that says if the mode was enabled or disabled.

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -31,7 +31,7 @@ asm4wb_icons_path = os.path.join( asm4wbPath, 'Resources/icons')
 global main_Assembly4WB_Icon
 main_Assembly4WB_Icon = os.path.join( asm4wb_icons_path , 'Assembly4.svg' )
 
-# import treeSelectionOverride as selectionOverride
+import treeSelectionOverride as selectionOverride
 
 
 """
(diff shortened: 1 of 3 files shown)


Zolko-123:
I think there was a question about such a toggle-switch in the forum, and it might not be possible to do. For sure it's not possible to change the icon.

Zolko-123:
I changed the icon and streamlined the function. I think it's actually very good, thank-you.

There is something that could be improved: allow this functionality not only for linked objects but Parts and Bodies. And also Fasteners.

abetis:
Fasteners are not linked parts?

Zolko-123:
> Personally I need that mode of selection in general working mode, not in the insertLink.

but it doesn't work for measures for example: if you want to measure the distance between de centres of 2 circles, the link selection mode is in your way. Enabling/disabling it works well, thank-you

Zolko-123:
no, fasteners are not linked parts.

abetis:
ok, got it

Zolko-123:
> 
> 
> Adding toolbar button to enable/disable select mode.
> Unfortunately it looks like there is no way to update the icon after addition of the command on initialization.
> All I could do is add a printout that says if the mode was enabled or disabled.

There is a way now, see 
https://forum.freecadweb.org/viewtopic.php?f=23&t=55958&p=481531#p481531

PR #137, opened 2020-11-06
placeLinkCmd: Specify AttachmentOffset (translation and rotation) of linked part as absolute values

DerAndere1:
This is targeted at the master branch. If you prefer a pull request targeted at the development branch, you can merge https://github.com/Zolko-123/FreeCAD_Assembly4/pull/138 instead.

This adds SpinBoxes for **specifying absolute AttachmentOffset values (translation and rotation)** in the "Place Linked Part" dialog.
When opening the dialog, the default values of these SpinBoxes correspond to the original AttachmentOffset of the linked part.

Picture2

The rest is similar to my previous pull request https://github.com/Zolko-123/FreeCAD_Assembly4/pull/136 (which is outdated and closed): 

Now if two LCS were selected in the tree view, those LCS are pre-selected as attachment-points in the "Place linked Part" dialog.

**Usage:** 

To Move/attach a linked part in an Assembly4 assembly, you now have two options:

workflow a): 

1. Select the App::Link for the linked part
2. Clicking the button "Edit Placement of Part" (Tooltip: "Move/Attach a part in the assembly")
3. Optionally, change the value of the SpinBoxes to **specify a rotation angles and ** (range: -360.00 to 360.00, default: taken from the original AttachmentOffset of the linked part)
4. Optionally, left-click on the button "Rot X + 90°", "Rot Y + 90°" to rotate the AttachmentOffset of the linked part in 90°-steps. The values are taken from the SpinBoxes (see in step 3).

workflow b):

1. In the model tree, first select an LCS of the linked part
2. Ctrl + left-click to add to the selection another LCS (in the parent_part) that you want to use as the attachment point
3. Left-click button "Edit Placement of Part" (Tooltip: "Move/Attach a part in the assembly") to open the dialog "Place Linked Part"
4. Observe that the two LCS that were selected in step 2 and step 3 are also preselected in the dialog
5. Optionally, change the value of the SpinBoxes to **specify a rotation angles and ** (range: -360.00 to 360.00, default: taken from the original AttachmentOffset of the linked part)
6. Optionally, left-click on the button "Rot X + 90°", "Rot Y + 90°" to rotate the AttachmentOffset of the linked part in 90°-steps. The values are taken from the SpinBoxes (see in step 3).

**Related feature requests:**

- https://github.com/Zolko-123/FreeCAD_Assembly4/issues/64
- https://github.com/Zolko-123/FreeCAD_Assembly4/issues/85
- https://github.com/Zolko-123/FreeCAD_Assembly4/issues/78

Issue #85, opened 2020-06-29
[enhancement] Allow to spefify how to rotate in placement dialog


donovaly:
When placing a part, one gets a dialog with 3 buttons "Rot X" etc. These rotate always 90°. But there are (not so rare) cases where one needs a different angle. Take e.g. this example:
![FreeCAD_bOnIHH42kI](https://user-images.githubusercontent.com/1828501/86042696-d52a8780-ba47-11ea-8774-0c9b754479ae.png)

Here I need at first a rotation by 90° around the y-axis, then subsequently a rotation of 30° around the z-axis.

I know this is a bit tricky because as it currently, one can only rotate around one axis. 

I thought about it, and I think now the solution would be to add to the placement dialog the attachment part of FC's datum/primitives dialog. This would allow to rotate around several axes and also fix issue #78. But let me think about this more.

Zolko-123:
I agree with #78 but not on this one.

Issue #78, opened 2020-06-25
[enhancement] dialog for setting attachment


donovaly:
After placing a part one often needs to change the attachment relative to the LCS. This has to be done in the Data tab, section Assembly -> Attachment.

However, it would be nice if there would be a dialog to do this. I think A4 could use the existing Edit Datum dialog for this task.

Zolko-123:
you mean: after attaching a part to an LCS in the assembly, you would want to change the AttachmentOffset directly in the placement dialog ?

donovaly:
I would like to right-click on the part in the tree and in the context menu there should be under "Assembly" an entry "Edit Attachment". This should open the attachment dialog. I mean the same dialog that I already get when double-click on an LCS in the tree:
![FreeCAD_WoblS42oxC](https://user-images.githubusercontent.com/1828501/85864275-9a64ed00-b7c4-11ea-8aa0-707c8694ccb9.png)

There I can easily and especially immediately visibly change the attachment. The reason is that it is often quicker not to add another LCS but to use an existing LCS and attach to this with an offset.

Zolko-123:
> right-click on the part in the tree

but Parts don't have attachment, only datum objects have them ! You cannot open that UI for parts and bodies. Actually, datums, sketches, Part features have attachment (MapMode)... but not Parts, Bodies, and Links: they don't have Attachments, only Placements.

donovaly:
> You cannot open that UI for parts and bodies.

Not yet. Therefore I labeled this issue as enhancement. I want to have this feature added to assembly4. It is not necessary to make this a feature for other WBs.

Zolko-123:
Unfortunately, this is a core FreeCAD functionality, nothing that Asm4 can do about. I close this issue

donovaly:
> Unfortunately, this is a core FreeCAD functionality, nothing that Asm4 can do about. I close this issue

I don't understand. The attachment dialog exists in FC. Why can't one call it whenever one likes?
The dialog is the file
https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Part/AttachmentEditor/TaskAttachmentEditor.ui
and it can be called via Python (I assume) using this file:
https://github.com/FreeCAD/FreeCAD/blob/master/src/Mod/Part/AttachmentEditor/TaskAttachmentEditor.py

So since it is part of the Part WB there should be a way to use it. Maybe I have a thinko. Please tell me.

[...]

Issue #64, opened 2020-05-13
[FR] preselecting LCS in 3D view for changing attachment


DerAndere1:
I would like to be able to change attachment of subassemblies/parts to the parent Model using the following workflow: 
1. Make visible all LCS of the assembly
2. Select one LCS of the subassembly and one LCS of the parent Model
by holding down Ctrl while clicking on both LCS **in the 3D view**
3. Click on the button "Move/Attach Part in the assembly". This should open the dialog "Place linked Part", but with those two LCS **preselected**.

**Usecase:**

Choosing attachment LCS visually in 3d view. Especially helpful for assembling parts with many LCS where choosing from a list of LCS becomes guesswork

DerAndere1:
I have no FreeCAD sripting experience but here is some rough pseudocode proposal for the "Place linked Part" dialog: 
```py
selection = FreeCADGui.Selection.getSelectionEx() # see https://wiki.freecadweb.org/Selection_API
sel_subobj_part0=selection[0].SubObjects # TODO: Find documentation about getSelectionEx()[0].SubObjects
sel_subobj_part1=selection[1].SubObjects

selected_LCS_part0 = [sel_SubObj for sel_SubObj in sel_subobj_part0 if (type(sel_SubObj) is PartDesign.CoordinateSystem)] # TODO: Is the type of an LCS really PartDesign::CoordinateSystem and what is the correct Python syntax?
selected_LCS_part1 = [sel_SubObj for sel_SubObj in sel_subobj_part1 if (type(sel_SubObj) is PartDesign.CoordinateSystem)] 

# Existing code for the "Place linked Part" dialog continues here ... . At the end of that code insert something like this:

place_dialog.LCS_part0_ListWidget.setCurrentIndex(selected_LCS_part0[0].Name) # TODO: Replace this with actual code that preselects the correct element from the left list widget in the "Place linked Part" dialog.
place_dialog.LCS_parentAssembly_ListWidget.setCurrentIndex(selected_LCS_part1[0].Name) # TODO: Replace this with actual code that preselects the correct element from the right list widget in the "Place linked Part" dialog.
```

Zolko-123:
yes, this would be a good feature, but I don't know how to implement it.

DerAndere1:
[...]

PR #138, opened 2020-11-08
placeLinkCmd: Specify absolute AttachmentOffset (translation and rotation) of linked part, LCS preselection

DerAndere1:
This is targeted at the development branch. If you prefer a pull request targeted at the master branch, you can merge https://github.com/Zolko-123/FreeCAD_Assembly4/pull/137 instead.

This adds SpinBoxes for specifying absolute AttachmentOffset values (translation and rotation) in the "Place Linked Part" dialog.
When opening the dialog, the default values of these SpinBoxes correspond to the original AttachmentOffset of the linked part.

Picture2

Also, if two LCS were selected in the tree view, those LCS are  now pre-selected as attachment-points in the "Place linked Part" dialog.

Usage:

To Move/attach a linked part in an Assembly4 assembly, you now have two options:

workflow a):

1. Select the App::Link for the linked part
2. Clicking the button "Edit Placement of Part" (Tooltip: "Move/Attach a part in the assembly")
3. Optionally, change the values of the SpinBoxes to specify translation and rotation (angle range: -360.00 to 360.00, default: taken from the original AttachmentOffset of the linked part)
4. Optionally, left-click on the button "Rot X + 90°", "Rot Y + 90°" to rotate the AttachmentOffset of the linked part in 90°-steps. The values are taken from the SpinBoxes (see step 3).

workflow b):

1. In the model tree, first select an LCS of the linked part
2. Ctrl + left-click to add to the selection another LCS (in the parent_part) that you want to use as the attachment point
3. Left-click button "Edit Placement of Part" (Tooltip: "Move/Attach a part in the assembly") to open the dialog "Place Linked Part"
4. Observe that the two LCS that were selected in step 1-2 are also preselected in the dialog
5. Optionally, change the value of the SpinBoxes to specify translation and rotation (angle range: -360.00 to 360.00, default: taken from the original AttachmentOffset of the linked part)
6. Optionally, left-click on the button "Rot X + 90°", "Rot Y + 90°" to rotate the AttachmentOffset of the linked part in 90°-steps. The values are taken from the SpinBoxes (see step 5).

**Related feature requests:**

- https://github.com/Zolko-123/FreeCAD_Assembly4/issues/64
- https://github.com/Zolko-123/FreeCAD_Assembly4/issues/85
- https://github.com/Zolko-123/FreeCAD_Assembly4/issues/78

### DIFF
--- a/libAsm4.py
+++ b/libAsm4.py
@@ -161,6 +161,40 @@ def getLinkAndDatum():
     return retval
 
 
+# get from the selected datum the corresponding link
+def getLinkAndDatum():
+    retval = (None,None)
+    # only for Asm4 
+    if checkModel() and len(Gui.Selection.getSelection())==1:
+        parentAssembly = App.ActiveDocument.Model
+        # find all the links to Part or Body objects
+        childrenTable = []
+        for objStr in parentAssembly.getSubObjects():
+            # the string ends with a . that must be removed
+            obj = App.ActiveDocument.getObject( objStr[0:-1] )
+            if isLinkToPart(obj):
+                # add it to our tree table if it's a link to an App::Part ...
+                childrenTable.append( obj )
+
+        selObj = Gui.Selection.getSelection()[0]
+        # a datum is selected
+        if selObj.TypeId in datumTypes:
+            # this returns the selection hierarchy in the form 'linkName.datumName.'
+            selTree = Gui.Selection.getSelectionEx("", 0)[0].SubElementNames[0]
+            (parents, toto, dot) = selTree.partition('.'+selObj.Name)
+            link = App.ActiveDocument.getObject( parents )
+            if dot =='.' and link in childrenTable:
+                retval = (link,selObj)
+            else:
+                # see whether the datum is in a group, some people like to do that
+                (parents2, dot, groupName) = parents.partition('.')
+                link2 = App.ActiveDocument.getObject( parents2 )
+                group = App.ActiveDocument.getObject( groupName )
+                if link2 in childrenTable and group.TypeId=='App::DocumentObjectGroup':
+                    retval = (link2,selObj)
+    return retval
+
+
 # get all datums in a part
 def getPartLCS( part ):
     partLCS = [ ]
--- a/placeLinkCmd.py
+++ b/placeLinkCmd.py
@@ -48,7 +48,7 @@ class placeLinkCmd():
 
     def IsActive(self):
         # We only insert a link into an Asm4  Model
-        if App.ActiveDocument and Asm4.getSelection():
+        if (App.ActiveDocument and (Asm4.getSelection() or (all(Asm4.getLinkAndDatum2())))):
             return True
         return False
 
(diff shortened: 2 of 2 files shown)


Zolko-123:
Thank-you, I'll test it now

Zolko-123:
Yes, I like it very much. Especially the translation, not so much the rotation: on a small-ish screen, you don't see the rotation buttons. And rotations with arbitrary values should be done with dedicated attachment LCS. What do you think ? Do you use very often this precise rotation in the placement ? I think it's simpler and cleaner like this :

placeLink

didn't test the pre-selection of LCS yet.

Zolko-123:
as for the pre-selection of LCS, I don't understand how it can work: if you want to attach a link to another link, there is no way to know which one is the link to be placed and which one is the target attachment. Also, there is no getLinkAndDatum2 in this PR, I get an error.

I was rather thinking about a selection observer, where it would be possible to select the LCS in the 3D window, rather than in the tree.

DerAndere1:
The free setting of rotation can be used in cases like a rotating turret where the top-part is connected to a platform in a way that allows rotation but where extrusions from both parts interfere with each other and thus limit the rotation to a small area. The same top-part may rotate freely on another platform, thus the correct rotation of the attachment LCS in the top-part cannot be inferred from the top-part alone. 
On the other hand, I agree that limiting the possibilities for the user to fool around with the rotation in the dialog might enforce a well-thought LCS placement (interface). 

I just saw that by accident I added a duplicate of the function getLinkAndDatum. I rather wanted to

add (taken from my previous PR) the following in libAsm4:

```
# get from two selected datums the corresponding links
def getLinkAndDatum2():
    retval = (None, None, None, None)
    # only for Asm4 
    if checkModel() and len(Gui.Selection.getSelection()) == 2:
        parentAssembly = App.ActiveDocument.Model
        # find all the links to Part or Body objects
        childrenTable = []
        for objStr in parentAssembly.getSubObjects():
            # the string ends with a . that must be removed
            obj = App.ActiveDocument.getObject( objStr[0:-1] )
            if isLinkToPart(obj):
                # add it to our tree table if it's a link to an App::Part ...
                childrenTable.append( obj )

        selObjA = Gui.Selection.getSelection()[0]
        selObjB = Gui.Selection.getSelection()[1]
        # two datum objects are selected
        if ((selObjA.TypeId in datumTypes) and (selObjB.TypeId in datumTypes)):
            # this returns the selection hierarchy in the form 'linkName.datumName.'
            selTreeA = Gui.Selection.getSelectionEx("", 0)[0].SubElementNames[0]
            selTreeB = Gui.Selection.getSelectionEx("", 0)[0].SubElementNames[1]
            (parentsA, totoA, dotA) = selTreeA.partition('.'+selObjA.Name)
            (parentsB, totoB, dotB) = selTreeB.partition('.'+selObjB.Name)
            linkA = App.ActiveDocument.getObject( parentsA )
            linkB = App.ActiveDocument.getObject( parentsB )
            if (dotA =='.' and linkA in childrenTable) and (dotB =='.' and linkB in childrenTable):
                retval = (linkA, selObjA, linkB, selObjB)

            else:
                # see whether the datum objects are in a group, some people like to do that
                (parentsA2, dotA, groupNameA) = parentsA.partition('.')
                (parentsB2, dotB, groupNameB) = parentsB.partition('.')
                linkA2 = App.ActiveDocument.getObject( parentsA2 )
                linkB2 = App.ActiveDocument.getObject( parentsB2 )
                groupA = App.ActiveDocument.getObject( groupNameA )
                groupB = App.ActiveDocument.getObject( groupNameB )
                if ((linkA2 in childrenTable and groupA.TypeId=='App::DocumentObjectGroup') and (linkB2 in childrenTable and groupB.TypeId=='App::DocumentObjectGroup')):
                    retval = (linkA2,selObjA,linkB2,selObjB)

    return retval
```

While we are at it, I will research how to add a selection observer that deals with LCS. I may be able to borrow from the link observer that was added recently to Assembly4

Zolko-123:
yeah, I saw that. I replaced with the new function, and the selection of 2 LCS to infer the attachment isn't very intuitive. Are you sure you want to leave it in ? Anyway, I'll push my mods on your PR.

Thanx

DerAndere1:
Yes leave that in, you do not have to document it. But it allows the workflow described in the last comment of feature request #64. This is great for parts with a lot of LCSs that are best selected with the help of the 3D view. Also: as long as there is no possibility at all in FreeCAD to select two LCS from the 3D view, I cannot think of a good selection observer.

Zolko-123:
When selecting 2 LCS, why do you want to highlight (change the appearence) of the second object ? In the "normal" behaviour, the object to be placed is white transparent in order to recognise it in an assembly. This should also be the case in the "new" behaviour. What is the rationale to also change the appearance of the second object ?

DerAndere1:
changing the appearance of the second part is not intended. it is rather an artifact due to my brainless copy-paste-modify coding.

I guess it is enought to remove all lines related to selectedLinkB from the following section:

https://github.com/Zolko-123/FreeCAD_Assembly4/blob/480529208ce29ca1edb6c4603876783da50b8c43/placeLinkCmd.py#L284-L292

Zolko-123:
You duplicated ALL the code ?????? You can't be serious ... also, it doesn't work correctly: when I select 2 LCS, it does highlight them in the selection list, but it doesn't place the link to where it should go. I have to change some selections, and then go back again, to put the link where I initially selected.

Next time, please split your PRs into separate functionality: the translation is very good, but this selection thing is not. A selection observer in the 3D window for LCS would be much better. There already is a selection observer for the part : you don't have to select the attachment part in the list, you can select it in the 3D window, and it will select it in the list. The same functionality would be good for the LCS.

Zolko-123:
I remove this dual LCS selection thing. If you **really** need it, please make another PR, but without duplicating **everyting**, because it's unmaintainable.

DerAndere1:
We have the code in my repo in case we want to improve the preselection code. Feel free to take only the bits you like.

Zolko-123:
Nice ... but where is that code ? Couldn't find it

PR #147, opened 2020-11-26
'Asm4_cloneFastenersToAxes' command was added.

AlexUg:
This command allows to clone fastener for placing one to hole axes.
Steps:
1. create and place a fastener;
2. select the fastener and hole axes to which the clones of selected fastener should be attached;
3. run command.

Command will create clones of the selected fastener for each selected
hole axis and attach them.

### DIFF
--- a/FastenersLib.py
+++ b/FastenersLib.py
@@ -19,6 +19,8 @@ import libAsm4 as Asm4
 
 
 
+logger = App.Logger('Asm4_Fstnr')
+
 
 """
     +-----------------------------------------------+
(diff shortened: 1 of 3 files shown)


Zolko-123:
Nice, but please make pull-requests against the development branch, not master

Zolko-123:
I changed the cloneFastener() function so that it now creates an App::Link to the original fastener: I feel that implicitely, when you use this function, you want all resulting fasteners to be the same. Please check.

AlexUg:
Your solution is better.
But I found a bug - if you try to clone fasteners to linked parts/bodies (for example, two parts based on one document), then the command does not work correctly.
The problem is that Gui.Selection.getSelection() returns an object from the linked document and I don’t know yet how to determine in which App::Link to the part from linked document the selection is actually. I will try to find a solution soon.

Zolko-123:
I'm not sure to understand what you mean, but the command you're looking for is probably 
`Gui.Selection.getSelectionEx('', 0)` 
You can get inspiration from the Asm4_Measure file

AlexUg:
Yes, that's what I need.
I'm new in scripting for FreeCAD, so I might not know some things that others think are simple.
I asked a question on the forum (https://forum.freecadweb.org/viewtopic.php?f=10&t=52654), but there is no answer yet. And I tried to solve the problem in a slightly different way.

PR #148, opened 2020-11-28
'Asm4_cloneFastenersToAxes' command was improved.

AlexUg:
Fixed incorrect behaviour when selected differend linked objects based
on one part/body from other document.

### DIFF
--- a/FastenersLib.py
+++ b/FastenersLib.py
@@ -19,9 +19,6 @@ import libAsm4 as Asm4
 
 
 
-logger = App.Logger('Asm4_Fstnr')
-
-
 """
     +-----------------------------------------------+
     |               Helper functions                |
@@ -32,12 +29,12 @@ def getSelectionFS():
     # check that something is selected
     if len(Gui.Selection.getSelection())==1:
         obj = Gui.Selection.getSelection()[0]
-        if isFastener(obj):
+        if Asm4.isFastener(obj):
             selectedObj = obj
         else:
             for selObj in Gui.Selection.getSelectionEx():
                 obj = selObj.Object
-                if isFastener(obj):
+                if Asm4.isFastener(obj):
                     selectedObj = obj
     # now we should be safe
     return selectedObj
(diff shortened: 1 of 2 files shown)


Zolko-123:
It's good practice to try to put most of the stuff into libAsm4, but for Fasteners it's special, because the workbench might not be installed. Did you try to delete the fasteners WB and see what happens ? That's why there is a Fasteners_Dummy.py file, the presence of that WB is checked in InitGui.py: if it's not present, we load the dummy package to avoid error outputs.

Which means that everything related to fasteners should be in Fasteners.py. Can you make the change ?

AlexUg:
The functions 'cloneFastener', 'makeExpressionFastener' and 'placeFastenerToLCS' are not actually fastener-specific. Maybe it's better to change their names? Or is it still necessary to move to the 'FastenersLib.py'?

Zolko-123:
The problem is not with the name but with the 
`from FastenerBase import FSBaseObject`
in _libAsm4_. But it's actually only used in `isFastener()` therefore, moving that one back to _FastenersLib_ should be enough. For the others:
- `cloneFastener()` is generic and could be used globally, could be renamed cloneObject and stay in _libAsm4_

but

- `makeExpressionFastener()` is fastener specific
- `placeFastenerToLCS()` uses makeExpressionFastener therefore is also fasteners specific

so these should move back into _FastenersLib_, especially as `splitExpressionFastener()` is also there. I prefer to have self-contained files that use as little cross-calls as possible, and put into the generic Asm4 library only those functions that are really used everywhere. Some people would disagree, I know.

AlexUg:
You right -- `isFastener()` must be moved back to _FastenersLib_.
For the rest, at the moment the following proposals are:
- remove function `makeExpressionFastener` and use `makeExpressionDatum` instead one (these functions are very similar, but I replaced the code in `makeExpressionDatum` with `makeExpressionFastener` - it works in both cases: fasteners and datums);
- replace the function name `placeFastenerToLCS` with `placeObjectByOriginToLCS` and leave it in _libAsm4_ (there is no more fastener-dependent code in this function).

Zolko-123:
OK, nice. I propose `placeObjectToLCS()` though (shorter). Can you please make a PR ?

PR #152, opened 2020-11-30
Changes in the behavior of the object and fastener placement commands.

AlexUg:
To the task dialogs 'FastenersLibs.placeFastenerUI' and 'placeLinkCmd.placeLinkUI'
the ability to select HoleAxes in 3D View has been added.

Steps for Part:
1. choose a part;
2. invoke the operation 'Move / Attach a Part in the assembly';
3. in 3D Viwer, select HoleAxes in the part to be attached,
        and in the part to which you want to attach;
        the order of selection is not important.

Steps for Fastener:
1. choose a fastener;
2. invoke the 'Edit Placement of a Fastener' operation;
3. in 3D Viwer, select the HoleAxis in the part to attach to.

PS: When invoking an operation in 3D Viwer, the entire assembly is selected,
    wait for the selection was canceled and then perform the selection of HoleAxes.

PPS: When clicking on an axis, a part or other element in 3D View can be selected,
     it is necessary to select the axis (it may take several clicks).
     The selected axis should be highlighted in corresponding list in the task dialog.

PPPS: When selection is performed there is less problems if the display mode of objects is 'Wireframe'.

### DIFF
--- a/FastenersLib.py
+++ b/FastenersLib.py
@@ -47,6 +47,7 @@ def isFastener(obj):
         return True
     return False
 
+
 # icon to show in the Menu, toolbar and widget window
 iconFile = os.path.join( Asm4.iconPath , 'Asm4_mvFastener.svg')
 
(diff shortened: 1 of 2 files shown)


Zolko-123:
I wanted this behaviour since a long time. Thanx.

Zolko-123:
The first selection of the LCS in the target link is un-nice: I'll have a look at what can be done

Zolko-123:
I think I improved the selection mechanism: can you please update and check ?

AlexUg:
It's working very nice, thanks!

PR #175, opened 2021-03-06
Change animation-step-logic to QtTimer-based actions, rather than handling all calculations in loops.

Thowil:
This PR contains some improvements to animateVariable/the Animation Dialog.

Initially, the goal was to just fix a "freeze-crash" when accidentally trying to start an animation with unsuitable user defined values (e.g. RangeBegin<RangeEnd and Step<0).
Eventually however, the run*-functions were ported from explicit loops (in reaction to button presses) to a QtTimer-based approach in order to prevent possible delays (and sleep-calls) inside the GUI-loop.

Also, values should now be updatable while the animation is running.

Please note that I based this on "development" and did not squash in order to make it easier to follow the changes.
Just let me know in case you'd prefer otherwise.
Thanks! :-)

### DIFF
--- a/AnimationLib.py
+++ b/AnimationLib.py
@@ -79,6 +79,7 @@ class animateVariable():
         # Initialize States and timing logic.
         self.RunState = self.AnimationState.STOPPED
         self.reverseAnimation = False  # True flags when the animation is "in reverse" for the pendulum mode.
+        self.ForceGUIUpdate = False # True Forces GUI to update on every step of the animation.
         self.timer = QtCore.QTimer()
         self.timer.timeout.connect(self.onTimerTick)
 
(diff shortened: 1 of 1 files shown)


Zolko-123:
Hello, thank-you for your PR. I'll test it and come back

Thowil:
Perfect, just let me know if I can help in any way.
Thanks for all the work you're putting into this btw, it is really much appreciated!

Zolko-123:
nice, works, and the code looks indeed cleaner.

BUT: while we're at it, there are 2 features about this animation that have been asked by other users, and I wonder whether you could take this opportunity to implement them:

1) save the values of an animation, such that if called again the same values are presented to the user: the idea being that there are chances that the exact same animation will be called over-and-over again, and it would be convenient if the previous parameters are restored (variable, begin, end, step). The states that you introduced could be a good place to store these values. For the variable, a look-up in the existing variables list might be necessary.

2) display the current value above the slider, so that the user can see the current value. This can be useful if the animation is used to optimise or test some parameters in a range.

cheers

Thowil:
Sure!
That would be useful -- I'll get to it and will see what I can come up with.

Zolko-123:
actually, there is a third one:

3) save an animation to disk (mp4, gif, picture-to-picture ...). This would be very useful for demonstration purposes

PR #179, opened 2021-03-08
removed duplicate file

Florin-Popescu:
Hi,

On windows case sensitive file names are not enabled by default (at least on my machine). So when I tried to install Assembly4 via the Addon Manager it failed to load since the lowercase asm4wb_locator.py was replacing Asm4wb_locator.py. Removing the old file works for me.


Zolko-123:
good catch, I'll remove the file. But don't do PRs against the master branch, do them against the development branch please

Zolko-123:
I removed it by hand. Thanx for the tip

Issue #161, opened 2021-01-11
Animation dialog does not correctly store previous state


donovaly:
I don't know if this is a bug or missing feature, however, it would be very helpful if the animation dialog would state the previous state since at the moment you have to reset the value range on every call of the dialog.

Here is an example:
![koT4Pwzt8i](https://user-images.githubusercontent.com/1828501/104229464-9e98e500-544c-11eb-91e6-e90d2371c1f5.gif)

So you can see the float ran from 0 to 360 but after the reopening there is no range anymore. It stored the last value and put this to both range fields.
However, the state of the Loop option is correctly stored.

Zolko-123:
This has been asked before, I have to think how to implement it (a global variable may-be)

donovaly:
> This has been asked before, I have to think how to implement it (a global variable may-be)

Maybe you can use the way you store the current angle. In my example above you see then on reopening the animation dialog the angle 74. This was obviously the last angle position and it is still available after closing the dialog. So maybe you can save the range values the same way.

Zolko-123:
that value is the **current** variable value

donovaly:
> that value is the **current** variable value

Ah, OK. However, you already store the state of the Loop and Pendulum property, so it should be possible to store the values of the range the same way, no?

Alonso-JAMM:
Maybe a dictionary could be used to store the value, range, step, and other data of a variable. For example, the Variables object could have a property "App::PropertyPythonObject" that would store a dictionary containing the animation data related to all variables. So the dictionary would look like this:
`{"angle1": {"value": 74, "Range_Begin": 0, "Range_End": 360, "step":1}, 
   "angle2":{"value": 10, "Range_Begin": 5, "Range_End": 20, "step":1}}`

Then the data stored in the dictionary could be used by the animation window. This would be helpful in storing the animation data of multiple variables at the same time

Zolko-123:
This should be fixed in the recent merge of Thowil's PR

PR #183, opened 2021-03-12
Overhaul of Animation Dialog, make animation related configuration-values persistent

Thowil:
Hi Zolko,

here are the first two changes requested after my last PR (#175).
Actually having the variable displayed above the slider made some shortcomings with slider handling more obvious.
I hence decided to give the dialog a small handling-overhaul.
Furthermore, all Variables will now have their associated configuration stored directly in the doc-file (using an additional property; basically as suggested in #161, which should be fixed with this PR.)

Just let me know if you have comments and whatever else might be needed to get this merged. :-)

I'll move ahead with animation-recording as suggested in your last post here:
https://github.com/Zolko-123/FreeCAD_Assembly4/pull/175#issuecomment-792345200=

Cheers!

PS: Again, PR based on development, non-squashed for details.

### DIFF
--- a/AnimationLib.py
+++ b/AnimationLib.py
@@ -194,6 +194,7 @@ class animateVariable():
     def setVarValue(self,name,value):
         setattr( self.Variables, name, value )
         App.ActiveDocument.Model.recompute('True')
+        self.variableValue.setText('{:.2f}'.format(value))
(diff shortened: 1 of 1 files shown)


Zolko-123:
Whia, this is great, works perfectly. Thank-you

Issue #173, opened 2021-02-18
Add translations to A4


luzpaz:
https://wiki.freecadweb.org/Translating_an_external_workbench
https://github.com/kbwbe/A2plus/pull/416

Zolko-123:
Done in the development branch (future v0.10)

Zolko-123:
done in latest v0.10, please test and report  (though I don't know how to add/update this)

luzpaz:
@yorikvanhavre lets add it to crowdin ? + what's the external workbench author workflow to push/pull translations?

yorikvanhavre:
Sure! If you want, @Zolko-123 I need your crowdin username, and I'll add you as admin to https://crowdin.com/project/freecad-addons then you can upload your ts file there

PR #187, opened 2021-03-26
Export animations

Thowil:
Hi Zolko,

this PR adds native mp4/animated gif as well as frame by frame export for animations. 
To add value when compared to standard recordings via a screengrabber or the like, I included some basic image compositing features 
(cf. https://youtu.be/7UXug_iR51U for a not-so-polished example). 
Everything is configurable via a dedicated dialog, accessed via a new "Export"-button in animateVariable. 
Btw: The exporter features a fairly simple interface, so it might be interesting for scripting aside from purely asm4-animations as well. 

Hope this is useful! 

Cheers, 
Thomas


Zolko-123:
this looks great. I had computer problems these last weeks so couldn't look into your proposal, but I wall do it ASAP in the coming days.

Zolko-123:
Hello, I have some questions:

1) why create an extra file for the AnimationProvider class, and why not implement it directly into AnimationLib ? That seems like the logical destination

2) why change lisbAsm4 with the resourcePath ? The motto is "don't fix if it ain't broke", therefore I prefer to not change the Asm4 library for such trivialities

3) what and why did you change the FreeCAD.png icon ? 

As a conclusion, could-you please reduce the amount of files modified to a minimum, in this case 2 (AnimationLib and AnimationExproter) ? Thank-you very much for this very important feature. And while we're at it, if this works well, I'm sure you will be asked to see whether we can combine this with some ray-tracing and rendering option, so please plan in advance, this could become a killer feature :-)

Thowil:
Hi Zolko,

thanks for your feedback -- I checked again and you're right, the path-join change in libAsm4 is indeed not needed.
I had some problems in the process with file-placement and thought that getting rid of the forward slash would help (I'm currently on a windows machine) but that now seems unrelated. Thus, the commit can be skipped.

I added the PNG-version of the icon as I thought it would be nice to have the logo-overlay enabled by default.
The currently checked in icon is in SVG format, which of course makes complete sense in Qt context. However, the animationExporter uses PIL for most of its work and PIL is not readily supporting SVG. Rather than creating additional conversion methods, I felt it made more sense to just add a PNG-version (at least for now) and see how to proceed with that later.

Lastly, I split the code into different files, as having it all in one file made that one rather lengthy. 
It also allowed to import the exporter-code only when the user actually requires it, which might make sense given that exporting to video/image is likely to be used way less than other features of Assembly4. 

This also has the positive side effect that the exporter can now be imported into other code (apart from AnimationLib).
Any class implementing `def nextFrame(self, resetAnimation) -> bool`  to set up the scene for the next frame can call the dialog and benefit from it (c.f. the macro below for a simple example of what I had in mind).

I have to add though that I realized that stylistically, my code was not really following general guidelines for python.
I have since cleaned it up to conform better with pep8 etc (really helps when using an IDE... :-)) Would you prefer me to also add the cleaned up versions here or in another PR maybe?

Cheers and thanks,
Thomas

PS: I also started to look into the translation framework as per #173 . Is that something you'd be interested to move forward with?
I believe that it could be gradually done (file by file or so whenever appropriate) as soon as there'd be agreement on the general concept/translation-markers for strings. 

-------

```python
import FreeCAD as App
import FreeCADGui as Gui

from AnimationProvider import animationProvider
from AnimationExportLib import animationExporter


# Code to setup each frame
class animTest(animationProvider):
    def nextFrame(self, resetAnimation) -> bool:
        self.framecnt = 0 if resetAnimation else (self.framecnt + 1)
        App.ActiveDocument.Sphere.Radius.Value += 1
        App.ActiveDocument.recompute()
        return self.framecnt >= 10


# Add object so that something is there to scale
App.ActiveDocument.addObject("Part::Sphere", "Sphere")
App.ActiveDocument.recompute()
Gui.updateGui()

# Use AnimationExporter with definition from above
aP = animTest()
aE = animationExporter(aP)
aE.openUI()
```

Zolko-123:
The Export window is quite messed-up on my computer:

Animation

Also, it's not very clear what options do what

Zolko-123:
This actually works !!!

anim

There are some bugs: 

- MP4 output doesn't work for me
- GIF output creates 2 loops when I ask for 1
- the File selection line widget has height scaling issues

A also removed anything shadow related (I commented out the code, it's still there)

Thowil:
Hi,

looking at the squashed controls, I noticed that it initially layouts fine on my system with a 4k screen, but I can still force the window to be too small for the widgets to make sense any longer.
From your screenshot it look as if the layout is broken for you from the time the window opens. Is that correct? 

Regarding MP4-output: I'll check with a fresh Win10 and see if some additional codecs etc are needed.
(I didn't have to install anything extra here when I wrote the code, but might have done so for other projects in th past.)
I might be able to check with Linux later, but don't have a suitable computer at hand for that right now.

Cheers!

PR #191, opened 2021-05-02
add object's label to proposed datum name

zisoft:
I like to give my LCS's descriptive names, especially when using lots of them. I use the object's label as part of the LCS name, i.e.:

"LCS_Bottom_FL" for the part "Bottom" on the Front Left
"LCS_Side_left_FR" for the part "Side_left" on the Front Right

This makes it much easier for me to see at once what LCS to use when assembling a part.

To automate this I would suggest to add the part's Label to the name of the new datum object on creation.

### DIFF
--- a/Asm4_Measure.py
+++ b/Asm4_Measure.py
@@ -991,7 +991,8 @@ class selectionObserver():
         global taskUI
         if pt1!=pt2:
             line = Part.makeLine( pt1, pt2 )
-            wire = App.ActiveDocument.addObject('Part::FeaturePython', name)
+            wire = App.ActiveDocument.addObject('Part::FeaturePython')
+            wire.Label = name
             wire.ViewObject.Proxy = setCustomIcon(wire, taskUI.lineIcon )
             wire.Shape = Part.Wire(line)
             wire.ViewObject.LineWidth = width
--- a/Asm4_Measure.py
+++ b/Asm4_Measure.py
@@ -991,8 +991,7 @@ class selectionObserver():
         global taskUI
         if pt1!=pt2:
             line = Part.makeLine( pt1, pt2 )
-            wire = App.ActiveDocument.addObject('Part::FeaturePython')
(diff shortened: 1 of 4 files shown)


Zolko-123:
This is a good idea, but there were reports by people using non-ascci characters in names that they are not correctly displayed. Did you try your modification with parts that have labels composed with Cyrillic or Chinese characters ? You can try with some of the following characters for example:

€£#&@çó%λ→αφ

zisoft:
I am aware of these naming problems but this has nothing to do with my modification. In fact, it has nothing to do with Assembly4, it is a character set issue within FreeCAD. One cannot set an object's label with such characters.

Zolko-123:
you cannot **name** objects with such character, but you can **label** them. So if you use the label of a part, and it contains such characters, and then you want to name an LCS with such characters, I suspect that some strange things could happen. Could you please test this ?

zisoft:
Confirmed. It is a problem within Assemby4's Label management

zisoft:
I think I have fixed these problems.
It is a problem of FreeCADs createObjects routines where a label with special characters is not handled correctly. You can easily reproduce it in the python console by simply:

    App.ActiveDocument.addObject('PartDesign::Body', '€€€')

This does not work properly, the label is messed up. This works fine:

    o = App.ActiveDocument.addObject('PartDesign::Body')
    o.Label = '€€€'

I have fixed this handling on all occurences of Assembly4, everthing works fine now.

Zolko-123:
yes, well, no, please don't mix several problems in one commit.

Zolko-123 (changes requested):
please restrict this commit to the file newDatumCmd.py and leave the others untouched

zisoft:
You are absolutely right, this had to be another branch. I have reverted the last changes.

Zolko-123:
Thank-you, both for the functionality and for listening to my requests

PR #199, opened 2021-05-23
Add .AVI with X264 codec as additional animation export format.

cyr123:
I was not able to export an animation to MP4 format on Ubuntu 21.04, this seems to be a known problem with cv2.VideoWriter

This simple cut-and-paste change adds the option of exporting to an .avi container instead, which as far as I can tell is the "safe" option with OpenCV video export.

### DIFF
--- a/AnimationExportLib.py
+++ b/AnimationExportLib.py
@@ -25,7 +25,7 @@ from AnimationLib import animationProvider
     +-----------------------------------------------+
     |               Animation Export                |
     | Provides ability to export the animation as   |
-    | single frames, mp4 or animated gif.           |
+    | single frames, mp4, avi or animated gif.      |
     +-----------------------------------------------+
 """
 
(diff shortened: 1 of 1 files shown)


Thowil:
Thanks for figuring out avi actually works in Ubuntu's case :-)
Interestingly enough, exporting into an avi-container fails for Windows (with otherwise identical codec -- 'X264' is the autoselected one for me).
If I see it correctly, this basically means that the only difference here is the export-filename, based on which the container format is selected.
Both functions could be merged back into one which then gets called for both extensions.
This still does not solve the problem that one format will apparently fail depending on the used OS, but at least direct-export to video would work for linux then as well...

Zolko-123:
> Both functions could be merged back into one which then gets called for both extensions.

this is great if you find a way to make this generic for all OSs. 

BUT: please don't make any PR against the _master_ branch, always against the _development_ branch. Therefore, I close this PR in the hope that you can make another PR that would work for all OSs (and that you make for the _development_ branch)

cyr123:
> > Both functions could be merged back into one which then gets called for both extensions.
> 
> this is great if you find a way to make this generic for all OSs.
> 
> BUT: please don't make any PR against the _master_ branch, always against the _development_ branch. Therefore, I close this PR in the hope that you can make another PR that would work for all OSs (and that you make for the _development_ branch)

Ah, sorry didn't pay attention to the branch structure.

I only implemented this to quickly scratch my own itch so to speak, but I will have a look at a better implementation if/when I have the time and assuming no one beats me to it.

PR #201, opened 2021-05-28
Fix Video Export for various container formats

Thowil:
This PR aims to fix video-export problems (file ending up empty) for animations.
VideoWriter will be called with specific fourccs and in case the export results in an empty file, the process will be repeated using another codec.
In case exporting fails with all given codecs, an error-message is written to the console.

Furthermore, avi-, mov- and mkv-containers are added to the list of supported video-files.

This should create valid files across different OSes, as long as any valid codec for the given container is available and listed.

### DIFF
--- a/AnimationExportLib.py
+++ b/AnimationExportLib.py
@@ -16,6 +16,7 @@ import pathlib
 import cv2
 
 import FreeCADGui as Gui
+import FreeCAD as App
 import libAsm4 as Asm4
 
 from AnimationLib import animationProvider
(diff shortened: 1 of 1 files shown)


Zolko-123:
Thank-you very much

Zolko-123:
Hi, this raises an error:

```
import cv2
13:06:01  ModuleNotFoundError: No module named 'cv2'
```

I think this means that OpenCV is not installed. Could you make an intercept of this error, such that if it's not present the "Export" button is inactive ? And/or output a warning message explaining that OpenCV must be installed ? (with `Asm4.warningBox('Text')`)

PR #203, opened 2021-05-29
Add basis for i18n

Thowil:
This PR demonstrates the setup needed for localization.

I mainly followed
https://wiki.freecadweb.org/Translating_an_external_workbench#Inside_each_FreeCAD_command_class:
but named the translation-function _atr instead to avoid more typing...

To check that the translation is actually working, I wrapped the relevant strings in a couple of files into the translations-functions and collected them using pylupdate (small helper script for local use included).
I then created a German translation of the extracted strings locally (using Qt's linguist) and released it to get the compiled qm-file used during runtime. 

Adding translation surely looks a bit messy, as it introduces a lot of changes scattered throughout many files.
Luckily enough, most changes are trivial, albeit a bit distracting in the code itself until one gets used to it...

I did not try out crowdin or the like (or any automation to upload the "vanilla" ts-file there, for that matter), but I'd assume that to be comparatively straight forward (e.g. using the freecad-addons crowdin (?)).


Zolko-123:
thanx, I'll merge this one with a big next update

Zolko-123:
Hello, I merged this, but ... I don't know what to think. I have some comments:

- you specify `pylupdate=pylupdate5.exe` to update the translation, but this is a Windows executable

- I have removed all AssemblyLib translation because your file was behind the main branch

- Assembly = Baugruppe ? Are you sure ? Shouldn't this rather be "Zusammenbau" ?

- in the .ts file, the lines of each message is specified:
```
<message>
    <location filename="../InitGui.py" line="294"/>
    <source>Assembly</source>
    <translation>Baugruppe</translation>
</message>
```
but when I add empty lines to the source file, thus changing the line number, this doesn't seem affected. Does the line number really need to be specified ?

Zolko-123:
also: I moved the translations folder to Resources/translations

PR #205, opened 2021-06-05
Allow animating variables from different documents

LHLaurini:
This is a short patch for animating variables of any document, instead of only `App.ActiveDocument`. This is useful for synchronizing values among multiple `Model`s. Seems to work well so far.


Zolko-123:
This is a good idea, nice, thank-you

PR #226, opened 2021-08-30
for Edit part Info and make BoM

FarmingSoul:
in Asm4_libs i just make change in the name of column
in info part i make a automatique script just for the steel profiles and in french langage
in makebom i kept checkModel but i make it compatible with the old and new and i use the information of edit part info for make a spreadsheet

### DIFF
--- a/Asm4_libs.py
+++ b/Asm4_libs.py
@@ -34,10 +34,12 @@ datumTypes = [  'PartDesign::CoordinateSystem', \
                 'PartDesign::Point']
 
 
-partInfo =[     'PartID',                       \
-                'PartName',                     \
-                'PartDescription',              \
-                'PartSupplier']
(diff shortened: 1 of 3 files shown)


FarmingSoul:
hello @Zolko-123 ,

I just see that : 

> Zoltan Hubert
> converted checkModel() to getAssembly() where needed 

I don't test it but i think he works for the old and new model.
so it will be re modified in makebom.

Have nice day

Zolko-123:
A few comments: 

- please don't modify *Asm4_libs* for the default **PartInfo** fields. Fields should be added and deleted through the _PartInfo_ UI
- can you please separate PRs for **BoM** and for **PartInfo** please ?

I'd like to accept first the **PartInfo** PR, and then the **BoM** PR please : a BoM makes only sense if we have a working PartInfo dialog

PR #239, opened 2021-09-18
clean and documenting code

FarmingSoul:
reload infoScriptInit but it's internal script


FarmingSoul:
-french translation-
if it's possible to remake ts file i will translate more.
thankx for all

Zolko-123:
Can you please separate the translation PR from the rest ? Also, please don't include InfoScript.py, I will not accept it.

FarmingSoul:
i @Zolko-123

InfoScript is a autofilling field for partInfo
I would like to make partInfo a configurable UI with auto-fillable fields and / or manual fields.
entering information is quite boring, especially since it is often automated.
moreover if the user only uses autofilled fields he does not have to take care of infoPart UI since makeBom can fill in the info all by himself !
What do you think about ?

PR #245, opened 2021-09-24
Handle import datum when multiple (3+) objects are selected in the tree.

abetis:
All will be imported with their default names.
Single and 2 objects selection were adapted to use the new functions, but the logic was not touched


Zolko-123:
yes, this is a good idea. 

BUT: selecting 2 elements will then be confusing for the user, therefore I suggest the following (in pseudo-code):

```
if len(selection==2):
  if selction [0] and selection[1] are Asm4.datumTypes:
    if selection[0] belongs to a child linked par and selction[1] belongs to self.rootAssembly:
      match the Placement of selction[0] to that of selection[1]
else:
  if all selection[i] is Asm4.datumType and all selection[i] belongs to self.rootAssembly:
    import the datum
```

Also, may-be should the function Asm4.getSelectionTree() take a Gui.getSelection() as input, rather than making it's own selection in the function.

What do you think ?
Does that make sense ?

Zolko-123:
Also: please make PRs against the development branch, I never merge into the master.

Also: Asm4 functions are called riceCooker() and not RiceCooker(), can you please adjust ?

abetis:
Strange, I thought the PR will be to development stream since I created mine from your development stream.
Looks like GitHub doesn't recognize it automatically.

Can you please explain what the 2 selections are supposed to do in words?
I've asked you about it on the issue page, but you didn't answer.

All my datum points are always on the main tree, so I don't see the use case you have.

Well, there is a big mixup with capitalization of function names...
GetResources               <- Capital first
IsActive                         <- Capital first
getSelectedDatums      <- low first
Activated                       <- Capital first

No problem to change my 2 functions, but consistency speaking, they should be capital first...

Zolko-123:
there are 2 use-cases for me:

1) import a datum, or several datums, from child parts into the root assembly: in this case, select as many datums as you want, all that is mandatory is that they all are in child parts of the root assembly, may-be deep, several layers down. This is useful to raise a datum from a part into a sub-assembly, to be able to attach the sub-assembly with some datum in a part. But this is also useful for imported STEP assemblies, where a part can be several layers down from the root assembly

2) set the position of an existing datum in the root assembly to the position of a datum in a child part. This is useful either to update the placement of a previously imported datum that has lost it's link, or if a datum had been used in the root assembly for construction, and later we want to match it to an existing datum in a sub-assembly (or part).

Does that make sense ?

(as for lowerCase and UpperCase, the way I understand is that in FreeCAD C++ function are UpperCase while Python function are lowerCase. But I'm not sure)

abetis:
I have all the code ready, moved the 2 sections special case to be handled separately, but I don't know how do you check the following from the data structures:
    if selection[0] belongs to a child linked part and selction[1] belongs to self.rootAssembly:
Specifically the "selection[0] belongs to a child linked par".
The "selction[1] belongs to self.rootAssembly" code is probably:
selction[1].getParentGeoFeatureGroup() == rootContainer
??

Zolko-123:
yes for selection[1]. 

For selection[0], we could use the Asm4.getSelectionTree() and check that [0]==rootAssembly.

abetis:
Moving to #249

PR #249, opened 2021-09-28
Multiple datum import

abetis:


abetis:
Please take a look.
If you have a file to test the special 2 selection case, I'll be happy to check it.
Can't test that code without an example.

Zolko-123:
`if len(selection)==2 and selection[0]==rootContainer and selection[1].getParentGeoFeatureGroup() == rootContainer:`

well, selection[0] will not be the rootContainer. You need to do:

```
( obj, tree ) = Asm4.getSelectionTree()
if tree[0]==rootContainer.Name :
or:
if App.ActiveDocument.getObject(tree[0])==rootContainer:
```

this should work because inside _getSelectionTree()_ only the first selected object is considered, which is what we want here

abetis:
Like that?

Zolko-123:
actually ... rootContainer is defined as `rootContainer = App.ActiveDocument.getObject(selTree[0])` some lines before, therefore you can drop that part of the test, and just leave:

`if len(selection)==2 and selection[1].getParentGeoFeatureGroup() == rootContainer:`

abetis:
Added the change (hope I didn't mess up with the branch tree with those rebases).
I don't like to deliver code that I can't test.
Please send a file for the special case test.

PR #250, opened 2021-09-28
add unit spin box

### LINKED ISSUE #243: Part Placement Units Assume Metric
When placing a linked part, the X, Y, and Z Translation units are in mm, regardless of whether Imperial units are active in FreeCAD. This differs from how most length inputs work. For example, the Pad operation in Part Design assumes units that match the Preference units, and allows for overriding those unites by typing "1mm" or 1in".

FreeCAD version 0.19-22039, Assembly4 version 0.10.4

xoviat:
fixes #243

### DIFF
--- a/Asm4_Measure.py
+++ b/Asm4_Measure.py
@@ -529,7 +529,7 @@ class selectionObserver():
         self.Pt2  = None
         PtS       = None
 
-    def render_distance(self, distance: int) -> str:
+    def render_distance(self, distance: float) -> str:
         return App.Units.schemaTranslate(
             App.Units.Quantity(str(distance) + " mm"),
             App.Units.getSchema(),
--- a/Asm4_libs.py
+++ b/Asm4_libs.py
@@ -813,5 +813,31 @@ def getSelectedDatum():
     return selectedObj
 
 
+"""
+    +-----------------------------------------------+
+    |                 Unit Spin Box                 |
+    +-----------------------------------------------+
+"""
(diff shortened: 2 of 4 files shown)


Zolko-123:
If imperial units are to be implemented, then it should be done using the user preferences settings, not adding new user preferences inside existing commands.

xoviat:
@Zolko-123 This uses the existing units in user preferences. I also re-implemented the option using a higher-level API. This is what this looks like now:

image

If you re-open this PR, I would be happy to discuss this to your liking, and I will commit to maintain this code (though I don't think changes will be needed for years).

Zolko-123:
hum, well, why not. BUT:

1) no cross-referencing between commands, therefore please put the _QUnitSpinBox()_ into the Asm4_libs.py.
2) what do you see in the stock parameter view of the FreeCAD UI about the AttachmentOffset ? Do the entry fields have units ?

xoviat:
Is this what you're looking for?

image

Zolko-123:
yes, thanx.

xoviat:
note: I did not put those units there. They were already present.

Zolko-123:
this breaks the entry fields for some locale settings, please see:

https://github.com/Zolko-123/FreeCAD_Assembly4/pull/252#issuecomment-932385798

So I reverted it in v0.11.1 in _FastenersLib_ and _placeLink_ to use `QtGui.QDoubleSpinBox()` (which seems to handle locale correctly) and not `Asm4.QUnitSpinBox()`.

PR #252, opened 2021-10-01
v0.11.0

Zolko-123:

PR #246, opened 2021-09-24
super infoPart and makeBom

FarmingSoul:
the infoPart all configurable with manuel field and auto filling field.
and there is a manage for the future add of new auto-field in Asm4 update.
you can use directly BoM without worrying about filling in the partInfo fields.
I write only in english and i comment the code and i clean it .
Try a commit in freeCAD before read the code in detail.

PR #253, opened 2021-10-01
Newbom

FarmingSoul:
use json for userconfig
make description for all i bellieve
infopartUi make user config
autofilling fonction....


FarmingSoul:
start without BoM and PartInfo on Part
0-start
I can make directly a BoM - all the auto filling field will be fill and a manual field stay blank. - BoM give me a Log
1-direct
and when i press OK he directly emene me in my BoM Spreadsheet
2-direct-ok
i can open Edit Part Info
3-epi
i can open config in Edit Part Info
4-epi-config
in config in Edit Part Info i can change name of field - active and disactive all field - add and delete manual field
5-epi-config-add-delete
when i press ok the config it save on file
6-epi-config-ok
you can see a new BoM with new config 
7-newbom-with newconf
when we update Asm4 with new autofilling field that was not before (see on the right(our update file) "newupdateofAsm4" is not on the left(confUserFile) --> Now is Json file in app_data_dir)
8-newautofieldinasm4update
so Edit Part Info - Config can see a new autofield and
the user can add our new autofilling field in Edit Part Info - Config
9-epi-config-update

Zolko-123:
This looks really great, thank-you. I can't merge it now but will look at it tomorrow.

Can-you please check the problem you mentioned with the latest 0.11.0 update ? https://github.com/Zolko-123/FreeCAD_Assembly4/pull/252#issuecomment-932385798

Zolko-123:
Nice, thank-you !!!! There was a small bug in InfoKeys:

```
name 'InfoKeys' is not defined
Traceback (most recent call last):
File "<string>", line 106, in Initialize
File "/home/hubertz/.FreeCAD/Mod/Assembly4_GitHub/infoPartCmd.py", line 20, in <module>
import InfoKeys
File "/home/hubertz/.FreeCAD/Mod/Assembly4_GitHub/InfoKeys.py", line 34, in <module>
for prop in InfoKeys.partInfo:
```

I also changed the user conf directory to "Templates" (~/FreeCAD/Templates)

Zolko-123:
BUT: there still is something that I don't like : that when making a BOM, it fills the part info for all parts (links). I'm not sure to have understood everything, but please remove the cross-references between InfoPart and makeBOM. 

What I suggest is the following:

1) incorporate InfoKeys into InfoPart
2) remove imports for them in makeBOM
3) in makeBOM, try to import the user conf JSON file
- if it doesn't succeed, use some pre-defined values from makeBOM
4) in makeBOM, don't use the create() and fill() methods when a property is not found, simply skip them A BOM should be passive, all it should do is to parse the assembly, when it finds a part it writes the info into the BOM. 

Actually, what we could do is to separate the part-info and BOM templates. The part can have much more info than the BOM needs, like supplier, mass, ...

Zolko-123:
Also, could you please have a look at the Configurations spreadsheet ? In the Assembly menu, select "open configuration panel" and create a new configuration. You'll see that the first cell is _A1 = 'Asm4::ConfigurationTable'_ which is a good signature to find it later. Could you please make the same layout for the BOM ? `A1 = 'Asm4::BillOfMaterials' ` , `A2 = Description`, ...

Zolko-123:
I made some updates that incorporate these changes, please update

Zolko-123:
@FarmingSoul : looking inside the code, on top of what I've said before, the _drawUI()_ function must only do that: draw the UI elements, and NOT do any other function (like 

```
        if self.infoTable[0][1]=='':
            self.infoDefault()
            self.addNew()
```

Also, in _infoPartConfUI()_, please call that _drawUI()_ also (and not _drawConfUI()_), and separate the logic from the UI please, this in order to make the code maintainable in the long term.

Zolko-123:
from https://github.com/Zolko-123/FreeCAD_Assembly4/pull/246#issuecomment-933352564:


> I can't reproduce the bug about not stored values. Just my new files are working good. I can't figure out what I've done, but in the old file with the bug I've found the redundant fields. May be this is the problem.

yes, I think that that's the problem. @FarmingSoul : I think that this should be checked when creating new properties, i.e. before creating a property check that it doesn't exist already (

```
if not hasattr(obj, property):
  create property
else:
  pass
```

FarmingSoul:
wow you have past lot of time on it !
Thankx it so cool

I have to take my time to translate and understand what you say and after i have to respond , wow, many hours but thankx for all, i learn a lot.

PR #256, opened 2021-10-02
fix unit spin box

xoviat:
this version should be more robust. I apologize for not catching the first error.

currently pending verification

### DIFF
--- a/Asm4_libs.py
+++ b/Asm4_libs.py
@@ -825,23 +825,20 @@ self.YtranslSpinBox = QtGui.QDoubleSpinBox() → self.YtranslSpinBox = Asm4.QUni
 class QUnitSpinBox(QtGui.QDoubleSpinBox):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.setSuffix(" " + self.get_default_unit())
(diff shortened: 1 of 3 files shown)


xoviat:
tested, but I'm going to leave this open for a bit to think of other things that might go wrong.

Zolko-123:
thanx, but I don't know how to test this.
.
.
.
.
therefore , since Qt seems to handle localisation correctly in its QDoubleSpinBox() class, can't we let Qt handle this here too ? Did you have a look at the QLocale() class:

https://doc.qt.io/qt-5/qlocale.html

xoviat:
The issue before was the string to float conversion failed because of localization.  This removes the conversion entirely and leaves rendering the float up to Qt. The QLocale isn't needed if there is no string conversion. I am almost certain this approach will work, but I am just going to think of how this could fail to avoid more headaches for you.

xoviat:
@FarmingSoul How do you change the locale for comma in units? Is it under windows preferences?

FarmingSoul:
@xoviat i think it in FreeCAD preferences

Capturexoviat

but may be windows ? i don't know

xoviat:
I finally got around to testing this. It appears to work with the french language settings:


image

PR #290, opened 2022-02-05
Updated Show and Hide LCSs Function

s-em-i:
Hi @Zolko-123,

please check these changes out.

In **InitGui.py** I've added **'Asm4_showLcs'** and **'Asm4_hideLcs'** to the Toolbar.
Is there a reason why it is sometimes `''` and mostly `""` ?

Not sure if this is the right place, but I've added in **Asm4_libs.py** a function called
**getDependenciesList**. It takes any kind of selection, marked in the Active Document.
And collects and creates a List of Objects. It doesn't return any Compounds, Bodys nor Linked Objects.
Those cannot contain any Asm4 Datum Types.

Finally in **showHideLcsCmd.py** I've changed the **IsActive** Sections to just check if there
is any kind of selection done or not.

I changed the **Activated** Sections as well. Now it takes any kind of Selection, doesn't matter
what was selected. The Objects we don't need will be ignored later in **showChildLCSs**.
It generates a List of Objects, and passes it to **showChildLCSs**.
Which will now take this List and hide/show every Datum Type.

It is very flexible, I hope so at least, and I tried it in quite a few combinations.
I hope this update will find positive receptions. Due to the fact, that hidden Datum Objects
get visible sometimes, after inserting a Link or a Part. I thought of a Hide LCS function and found one
already, "hidden", in your code. Thanks, but It didn't work so well, and it was limited to just one selection.
So I hope this does the trick now and you like it.

Cheers
Semi


Zolko-123:
looks nice, thank-you, I'll give it a try later

Zolko-123:
Hello, 

I hope I didn't do something stupid (I really don't know how to use git and GitHub), but please make separate PRs. 

While I like the show/hide functionality, why did you mix the ordering of things in placeLinkUI ? can you please remove that file from the PR ?

Zolko-123:
thanx

s-em-i:
Sorry, this wasn't meant to be in the PR. I don't know why this ended up there. I was just playing around with
the order, because I have to scroll quite long to get to the last added part.
But it only worked for one placement, and then the Select LCS in Parent: Window stays empty.

Zolko-123:
Hello @s-em-i , this doesn't seem to work for me. You can look at the attached assembly (the top-level file to open is asm-LaserCantilever). When I select (in the tree) asm_X or asm_Y, showLcs or hideLcs do the job with the original version, but don't do anything with the new one. It doesn't produce errors, but doesn't do anything.

Could you please have a look ? 

[Cantilever_v9.zip](https://github.com/Zolko-123/FreeCAD_Assembly4/files/8018806/Cantilever_v9.zip)

s-em-i:
I am sorry to hear that. It is working flawlessly on my machine. I am working now with your
development branch.
```
OS: macOS 10.15
Word size of OS: 64-bit
Word size of FreeCAD: 64-bit
Version: 0.19.24366 (Git)
Build type: Release
Python version: 3.9.7
Qt version: 5.12.9
Coin version: 4.0.0
OCC version: 7.5.3
Locale: C/Default (C)
```


https://user-images.githubusercontent.com/94402613/152876566-e49f7265-ca9f-40dd-9ca9-a1be616e89d8.mov
https://user-images.githubusercontent.com/94402613/152876578-9c4bbba5-4225-41ea-bce4-580f534e7edc.mov

Zolko-123:
you write this for getDependencies:

# Get almost all Objects within the passed Selection.
...
# Objects within Compounds and Bodys and also Linked Objects are left out.

what do you mean by "linked objects are left out" ?

s-em-i:
If the routine is looking for more possible objects within, it will eventually come to compounds and parts, which are linked
objects. If you would follow them up as well, python will break the loop after approx. 993 calls. So I figured to not do this,
due the fact, that any of your 4 datum types cannot be placed inside that levels anyway.

What do you mean by switch windows? If I click on a different sheet, FreeCAD shows me that window automatically.
If I have just the document name marked, this doesn't count as a selection.

But you are right, the asm_Mini4Wheel assy inside asm_X is not affected at all. Let me dig deeper into this...

Zolko-123:
I use to show a single document : when I select a sub-assembly in the assembly, I'd like that all datums inside that sub-assembly are shown/hidden with this function.

showLcs

s-em-i:
Yes me too. Weird, you have datums inside a Body. I thought this is not the way your Workbench works.

Zolko-123:
yes, you're right, but it's for compatibility reasons: some people do use datums in bodies, so ... when in Rome, do as the Romans.

I added the functions to the toolbar. Apart from that, is there a big advantage to use your version of show/hide LCS ?

s-em-i:
Well, I totally agree with you. I am trying to fix it...
The big advantage is the on one side the selection of more than just one object.
I have so many objects with datums that pop up all the time, it is just one click to hide them. Otherwise
I have look for them and click one by one.

s-em-i:
Perhaps this could fix it.
Change @ Line 328 in **Asm4_libs.py**
```
            #SubObjNames=Selection.getSubObjects()
            # Some Objects return None Objects,
            # even if it has the 'getSubObjects' attribute
            # 'getSubObjects' delivers uniqe Names with a trailing .
            for SubObjName in Selection.getSubObjects(1):              
                #SubObjects.append(App.ActiveDocument.getObject(SubObjName[0:-1]))
                SubObjects.append(Selection.getSubObject(SubObjName, 1))
```

Zolko-123:
If the question is about multiple selection, may-be the better solution is to extend the current function, which works well for single selections, to multiple selections. Rather than starting from skratch.

Zolko-123:
something like:

```
for s in Gui.Selection.getSelection():
  if s is link
  elif s is container
```

s-em-i:
> If the question is about multiple selection, may-be the better solution is to extend the current function, which works well for single selections, to multiple selections. Rather than starting from skratch.

True, I tried that first, but I couldn't get where wanted to be. So I started over.
I think the issue lies on me, I don't get all these different types and objects in FreeCAD.
Also, it seems there is some lack of API Documentation, which makes it even harder for me.
For example the function **getSubObjects**, If I hover over, it says:
`getSubObjects(reason=0): Return subname reference of all-subjects`
No explanation on different arguments or whatsoever.

I even tried different values than `()` or `(0)`, but result was always the same, so no differences.
Now with the issues you have reported back to me, I found out, that if you call it with `(1)`, it actually
returns the referenced object. And not `None`, if if I call it with `()` or `(0)`.

Another example is the **getSubObject** function. There is a much better explanation yes, but it doesn't
say why you have to get rid of the `.` in the name, if you are just looking with the name as an argument.
And why you not have to get rid of it, if you are changing/adding the second parameter.

Now with these two changes, it seems to work. If you don't mind, I am keeping your files for testing
other changes. It is nice to have a "worst-case" or a "real-life" scenario to test on.

Can you recommend a better source to get Python (FreeCAD specific) Informations?
I've found the [FreeCAD source documentation](https://freecad.github.io/SourceDoc/), but it is a nightmare
to search and find something readable in there. Even If I go through a search engine. It does show me results
which are related to my search, but If I open it, the whole page is just a mass and I cannot even find (strg+f) the
string pattern which I can clearly see in the results.

For example, If I search for `site:https://freecad.github.io/SourceDoc/ getSubObjects` there comes up a
result like this.
Screenshot 2022-02-08 at 21 39 35

If I open the link, and search for the mark, it says `not found`.
Screenshot 2022-02-08 at 21 41 12

If I search for just for the functions name, I get this:
Screenshot 2022-02-08 at 21 42 11
Screenshot 2022-02-08 at 21 43 08

I am not complaining, I am just saying how it is, and If you have something different I can check out?

Zolko-123:
yes, I have the same: getSubObject, getSubObjects(), getSubObjectList(1, True), getSubObjectListOfSomeThing ... it's a mess, and that's why I mostly use quite crude functions, and not the very smartest of them. It makes the code clumsier than it could be, but easier and more predictable to handle.

For our case, I'd try (but I will not have the time in the next days), 

```
isActive():
  if Gui.Selection.hasSelection() or Asm4.getAssembly():
   return True
```

and then, in Activated():

```
# if something is selected
if Gui.Selection.hasSelection():
 for sel in Gui.Selection.getSelection():
  if sel.isDerivedFrom('App::Link'):
    showChildLCSs(sel, False, processedLinks)
  elif sel.TypeId in Asm4.containerTypes:
    for objName in sel.getSubObjects(1):
          showChildLCSs(sel.getSubObject(objName, 1), False, processedLinks)
# if not, apply to the entire assembly
elif Asm4.getAssembly():
 assy = Asm4.getAssembly()
  for objName in assy.getSubObjects(1):
        showChildLCSs(assy.getSubObject(objName, 1), False, processedLinks)
```

well, you get the thing

s-em-i:
Ok, at least I am not the only one who is noticing this mess, thanks.

Did you tried my last correction? I am using it now since yesterday at it seems to work.
But I will check yours as well.

Zolko-123:
I updated the show/hide command for multiple selections, but using the old code. Seems to work well in my environment. It would be nice if you could test (the development branch) and tell whether that fulfills your needs

s-em-i:
> I updated the show/hide command for multiple selections, but using the old code. Seems to work well in my environment. It would be nice if you could test (the development branch) and tell whether that fulfills your needs

Thanks, I tried it. It does work with multiple selections yes. But it doesn't find as deep as before all the LCS objects.

PR #297, opened 2022-02-20
Select the first LCS if there is one

s-em-i:
Hi @Zolko-123,

If I place a part/link, pick a LCS from the **Parent Part** and forgot to pick a LCS from the **Linked Part**,
I will get the warning: **Problem in Selection**
Problem in Selection

So I figured It could make sense, to have the first LCS already selected, if there is one.
No Problem in Selection

Especially If I have to insert quite a lot of parts and they usually have only one LCS anyway.
It saved me quite some time, not to click all the time the first and only LCS and close the warning window.

This change does not effect the Move/Attach Function.

Cheers
Semi

### DIFF
--- a/placeLinkUI.py
+++ b/placeLinkUI.py
@@ -165,6 +165,11 @@ class placeLinkUI():
         if lcs_found:
             # ... and select it
             self.partLCSlist.setCurrentItem( lcs_found[0] )
+        else:
+            # If no LCS is selected, select the first LCS if there is one
+            firstLCSItem = self.partLCSlist.item(0)
+            if firstLCSItem is not None:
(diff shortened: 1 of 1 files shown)


Zolko-123:
I've thought about this too, thank-you for the code. I'll test ASAP

Zolko-123:
I changed the code so an LCS is pre-selected only if there is a single LCS present. It's in the new version 0.11.10, thank-you

s-em-i:
Thanks, this not exactly what I had in mind, because this will produce again the **Problem in Selection** warning,
even if they are more than 1 LCS present. So the `>=1` solution would get rid of this warning at all.

PR #300, opened 2022-03-17
Fix various typos

luzpaz:
Fix both source and source comment typos.


luzpaz:
Sorry about the noise. I made the PR against development but turns out in was in error. So closing in favor of #301

PR #301, opened 2022-03-17
Fix various typos

luzpaz:
Fixed both source and source comment typos.


AnimationExportLib.py
```diff
@@ -698,9 +699,9 @@ def __adjust(self):
 
 
     # grab the close-event and ensure cleanup
-    def closeEvent(self, evnt):
+    def closeEvent(self, event):
         self.owner.onClose()
-        super().closeEvent(evnt)
+        super().closeEvent(event)
```
luzpaz, line 704: I searched to see if `event` was used as a variable anywhere else in the code and couldn't find it. So, I was wondering if it would be OK with you to use `event` instead of `evnt` ?

Zolko-123:
please make PRs against the development branch next time

Issue #307, opened 2022-04-18
Show/Hide LCS does not work for items in groups


howie-j:
The Hide/Show LCS button does not work for objects that are placed in a group.

(I tried adding 'App::DocumentObjectGroup' to "Asm4_libs.py" containerTypes, but it still does not work.)

Correction, it works. Will work on a PR.

howie-j:
Thanks for merging :)

PR #308, opened 2022-04-18
Fix 'Hide/Show LCS buttons does not work for items in groups' issue

howie-j:
Fixes issue #307. Seems to work with groups, parts, bodies and links.


Zolko-123:
thanx. For the next time, please make the PR against the development branch

PR #311, opened 2022-04-24
PR#308 regression fix

howie-j:
I was not aware that many files and functions uses containerTypes, so reverted to before my last PR and check for groups in the if statement instead.


hasecilu:
Hello, I made the changes in both files in `.FreeCAD/Mod/Assembly4` but the behavior remains the same, there is another step to do?

howie-j:
> Hello, I made the changes in both files in `.FreeCAD/Mod/Assembly4` but the behavior remains the same, there is another step to do?

Hi hasecilu, this fix is for the error generated when creating new assemblies/parts/groups.

I have not made a PR to Zolko yet for the fix regarding including parts from groups in BOM, you can find it here:
https://github.com/howie-j/FreeCAD_Assembly4/commit/18d9483f8ddc34b5db081bda336e4de68e3278f9

- Regards, Håvard

Issue #323, opened 2022-08-12
The number of decimals in Place Linked Part should match the value in the global preferences


martinda:
In the Place Linked Part box, the number of decimals is currently 2, regardless of the global preferences (mine is set to 4). When using the imperial system, more decimals are needed. For example 3/8 of an inch is 0.375 inch. I am looking through the code to figure out how to fix this issue.

I see the `length_divisor` in the [`QUnitSpinBox`](https://github.com/Zolko-123/FreeCAD_Assembly4/blob/57b59e859120d735ce2225abb3ec7fa652f97ba2/Asm4_libs.py#L696), but I don't know if that is relevant or not.

Then I looked at possibly adding a call to [`setDecimals()`](https://srinikom.github.io/pyside-docs/PySide/QtGui/QDoubleSpinBox.html#PySide.QtGui.PySide.QtGui.QDoubleSpinBox.setDecimals) for each axis directly in the [`placeLinkUI.py`](https://github.com/Zolko-123/FreeCAD_Assembly4/blob/57b59e859120d735ce2225abb3ec7fa652f97ba2/placeLinkUI.py#L726) code, but the question is where to get the number of decimals in the first place.

In my search for where to get the number of decimals, I searched in [`UnitsApiPy.cpp`](https://github.com/FreeCAD/FreeCAD/blob/master/src/Base/UnitsApiPy.cpp) but I could not find a way to get the number of decimals.

Please let me know if I am on the right track or not, and let me know if you have any advice to help me fix it.

Zolko-123:
Thanks for the fix, will be in the next release

PR #324, opened 2022-08-12
Set number of decimals in UI to the global preference value

martinda:
I would like to propose this fix for issue #323 

Please let me know if this is the right approach. Let me know if there are other places where this change would be welcomed.


Zolko-123:
Thank-you, but please make PRs against the development branch.

PR #325, opened 2022-08-15
Allow rotation to be any degree from 0 to 360 (respecting the number of decimals)

martinda:
The current rotation when placing a link is only in multiples of 90 degrees. I need the ability to rotate at 45, so I thought I would make it a floating number instead, with as many decimals as specified in the application global configuration.

This PR builds on top of PR #324 for the variable `numberOfDecimals`.

Let me know what you think and thank you for creating this useful addon.

### DIFF
--- a/placeLinkUI.py
+++ b/placeLinkUI.py
@@ -516,25 +516,16 @@ class placeLinkUI():
         self.reorientLink()
         
     # Rotations
-    def onRotX(self):
-        if self.XrotationAngle > 270.0: 
-            self.XrotationAngle = self.XrotationAngle - 270.0
-        else:
-            self.XrotationAngle = self.XrotationAngle + 90.0
+    def onRotXValChanged(self):
+        self.XrotationAngle = self.RotXSpinBox.value()
         self.reorientLink()
 
-    def onRotY(self):
-        if self.YrotationAngle > 270.0: 
-            self.YrotationAngle = self.YrotationAngle - 270.0
-        else:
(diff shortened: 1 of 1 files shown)


martinda:
Oh I forgot to change `placePartUI.py`. Let me look at that.

martinda:
Sorry, I don't know what to do to in FreeCAD to get the code inside `placePartUI.py` to be executed. I tried adding a part from a separate file but that also called `placeLinkUI.py`. What causes `placePartUI.py` to be executed?

martinda:
This is buggy, the value does not appear to be saved when closing and re-opening the placement of the linked part.

martinda:
I was able to fix the saving and restoring of values, but I had to create PR #327 to fix the Z and X swap issue first.

martinda:
This is still buggy, the rotation value is not restored property after re-entering the "Edit Placement of a Part".

martinda:
I am starting to realize I do not have a proper understanding of the problem here. I am reading up on Euler angles and extrinsic and intrinsic rotations in Wikipedia. Sorry for all the noise I am creating.

Zolko-123:
The rotation by 90° serves to orient the part LCS w.r.t. the assembly LCS. So the goal is only to swap X, Y, Z orientations, thus only 90°. If you need a continuous rotation, create an intermediate LCS with the corresponding rotation and attach your part to it. The objective here is coding simplicity and robustness, at the expense of 1 added extra LCS.

PR #327, opened 2022-08-15
Fix X and Z rotation swap issue

martinda:
It appears to me that Z and X rotation values were swapped in `placeLinkUI.py`. Based on the [Yaw-Pitch-Roll documentation](https://wiki.freecadweb.org/Placement#Position_and_Yaw.2C_Pitch_and_Roll) found in the wiki, the mapping is as follows:

* Yaw - Z
* Pitch - Y
* Roll - X

Which means the value returned by `getYawPitchRoll()` is a tuple of degrees for the axis of (Z, Y, X) respectively - not (X,Y,Z).

In addition, the [`toEuler()`](https://github.com/FreeCAD/FreeCAD/blob/68053e4f4d75713887771db948bcd2d854d6f8cb/src/Base/RotationPyImp.cpp#L475) function appears to be an alias to `getYawPitchRoll()` function, so I changed the code to call it directly.

I guess `placePartUI` should be fixed but I don't see the UI button to invoque it.

### DIFF
--- a/placeLinkUI.py
+++ b/placeLinkUI.py
@@ -497,9 +497,9 @@ class placeLinkUI():
     def reorientLink( self ):
         moveXYZ = App.Placement( App.Vector(self.Xtranslation,self.Ytranslation,self.Ztranslation), self.old_LinkRotation )
         # New AttachmentOffset rotation of the link is difference between set rotation angles and original AttachmentOffset rotation of the link
-        rotationX = App.Placement( App.Vector(0.00, 0.00, 0.00), App.Rotation( App.Vector(1,0,0), self.XrotationAngle - self.old_LinkRotation.toEuler()[0] ))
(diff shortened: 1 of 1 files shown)


Zolko-123:
Is this a cosmetic fix or does it solve a real problem ?

martinda:
For the change from `toEuler()` to `getYawPitchRoll()`, I made it when I discovered that `toEuler()` was an alias to `getYawPitchRoll()`, so for me it was one less thing to try to remember as I was learning the code.

The change regarding the argument ordering was made because when I compared the code to the documentation it was confusing to me (there was a contradiction between the two). Maybe this contradiction is only cosmetic, or maybe I just don't understand.

Since my knowledge in this area is insufficient, I am leaving this PR up to you, and thank you for your help.

Zolko-123:
well, don't fix if it ain't broken, so I'll drop this PR for now. Thank-you for your attention though

PR #343, opened 2022-09-08
Case insensitive filter when inserting a part

leoheck:
Hey Zolko, this PR improves the Insert part with Case insensitive filter.

### DIFF
--- a/insertLinkCmd.py
+++ b/insertLinkCmd.py
@@ -324,13 +324,10 @@ class insertLink():
         filterStr = self.filterPartList.text().strip()
         for x in range(self.partList.count()):
             item = self.partList.item(x)
-            match_object =  re.search(filterStr, item.text(), flags=re.IGNORECASE)
             # check the items's text match the filter ignoring the case
-            if match_object:
-                if filterStr in item.text():
-                    item.setHidden(False)
-                else:
-                    item.setHidden(True)
+            match_object =  re.search(filterStr, item.text(), flags=re.IGNORECASE)
+            if filterStr and not match_object:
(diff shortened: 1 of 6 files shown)


Zolko-123:
That didn't work. This works, can you please check ? (and please don't mix whitespace cleanups with real code. Also, please do PRs against the development branch)

```
            match_object =  re.search(filterStr, item.text(), flags=re.IGNORECASE)
            if filterStr:
                if match_object:
                    item.setHidden(False)
                else:
                    item.setHidden(True)
            else:
                item.setHidden(False)
```

Zolko-123:
shorter even

```
            # check the items's text match the filter ignoring the case
            match_object =  re.search(filterStr, item.text(), flags=re.IGNORECASE)
            if filterStr and not match_object:
                item.setHidden(True)
            else:
                item.setHidden(False)

```

leoheck:
Yeah, I decided to not remove the old code... this filerStr test was already there. 
But I am testing this while using it a bit, it finds more things than it should find. 
I am thinking that the `item.text()` has extra data that is not being shown in the results. 
Did you try to use it? I am thinking I have to debug the results more.

leoheck:
> (and please don't mix whitespace cleanups with real code.

I did not do any whitespace cleanup.

leoheck:
Indeed, that is working better. I was still testing and the results were a bit wrong. Thanks, dude.

Zolko-123:
Thank-you, I copied the function by hand, (I didn't include all the " _Name :_ " modifications). This is now in the latest release

PR #352, opened 2022-09-11
New feature Expression Array

Jolbas:
I had an idea of a more flexible array where you can express the placement of its element as a function of its index. This way I could make more complicated patterns of elements. I did use the Polar Array code and come up with this solution. You fill in  an expression in ElementPlacement property and use ElementIndex as a part of the expression wherever you want to use the index of the child element.
I tried to attach an example file but failed. Here is an example of expression to use:
`create(<<placement>>; create(<<Vector>>; 0; 0; 50 + Variables.st_height * (floor(ElementIndex / 4) + 0.5)); create(<<rotation>>; 90 * ElementIndex; Variables.st_angle * ((ElementIndex + floor(ElementIndex / 4)) % 2 == 1 ? 1 : -1); 0))`
Maybe there are better ways to accomplish this


Jolbas:
It was possible to upload a zipped freecad document:
[parametrar.FCStd.zip](https://github.com/Zolko-123/FreeCAD_Assembly4/files/9543755/parametrar.FCStd.zip)

Zolko-123:
that's actually interresting ... better than my circular array because circular arrays are covered by Draft_Array (even though I don't like that function). So this ExpressionArray could be a nice touch. Can you give some expression examples how this could be used for linear arrays, circular arrays, spiral arrays (stairs), and/or other arrays that come to your mind ? 

We could even think about using this as a general array solution, with some common use-cases (see above) pre-written so the user can chose them for these common cases. And write his own for more exotic uses

Jolbas:
I will try to do some examples tonight. I see now that I uploaded the wrong file:
[pelarbit.FCStd.zip](https://github.com/Zolko-123/FreeCAD_Assembly4/files/9547780/pelarbit.FCStd.zip)

Jolbas:
A few examples

Spiral:
`create(<<placement>>; create(<<Vector>>; 0; 0; 40 * ElementIndex); create(<<rotation>>; 30 * ElementIndex; 0; 0); LCS_0.Placement.Base)`

Polar full revolution:
`create(<<placement>>; create(<<Vector>>; 0; 0; 0); create(<<rotation>>; 360 / ElementCount * ElementIndex; 0; 0); LCS_0.Placement.Base)`

Linear along Axis_1:
`create(<<placement>>; Axis_1.Placement * create(<<vector>>; 0; 0; 1) * 50 * ElementIndex; create(<<rotation>>; 0; 0; 0))`

Jolbas:
Four rows of elements in a wave :)
`create(<<placement>>; create(<<Vector>>; ElementIndex * 10; ElementIndex % 4 * 40; sin(ElementIndex * 10) * 60); create(<<rotation>>; 0; -cos(ElementIndex * 10) * 45; 0))`

Jolbas:
The expression in ElementPlacement tend to be long and hard to read and write. And I couldn't find a way to do the calculations in a spreadsheet. There's a cyclic reference if you use the ElementIndex from  the expression array.

The most flexible expression to create a placement has the form `create(<<placement>>; Base; Rotation; Center)`. Maybe we could pre-add a property for each of those arguments. Say we call them ElementBase, ElementRotation and ElementCenter and preset ElementPlacement to `create(<<placement>>; ElementBase; ElementRotation; ElementCenter)` Then the user just need to deal with ElementBase to accomplish a linear array.
It's of course possible for the user to right click in property editor and add custom properties too if reusable subcalculations are needed.

Zolko-123:
What I mean by that is that I'd like to use this ExpressionArray to **also** make circular and linear arrays easily

Zolko-123:
> Polar full revolution: 
> `create(<<placement>>; create(<<Vector>>; 0; 0; 0); create(<<rotation>>; 360 / ElementCount * ElementIndex; 0; 0); LCS_0.Placement.Base)`

There is something missing at the end:

`create(<<placement>>; create(<<Vector>>; 1; 0; 0); create(<<rotation>>; 360 / ElementCount * ElementIndex; 0; 0); LCS_0.Placement.Base) * SourceObject.Placement`

Jolbas:
ElementBase was a misleading name maybe. I'm not sure I fully understand the direction property. Is it not confusing to both have both X,Y,Z and direction? I agree we want the default state to be as simple ground as possible for the user to create linear and polar patterns and a combination of them. But we have to decide whether to default to XYZ displacement or direction+length. I most often prefer XYZ. Other directions can also be handled by the placement of the ExpressionArray.

What do you think of skipping Base,Rotation,Center properties and only have:
```
obj.addProperty("App::PropertyVector" , "LinearStep", "Array", '')
obj.addProperty("App::PropertyAngle" , "AngularStep", "Array", '')
```
And the rotation axis is in its default assumed to be z-axis
Then we set the default expression on ElementPlacement to
`create(<<placement>>; LinearStep * ElementIndex; create(<<rotation>>; AngularStep * ElementIndex; 0; 0)`

The user can then set AngularStep expression to `360 / ElementCount` if they want a full revolution.
They can set X,Y,Z in LinearStep or an expression on one of them to `200 / (ElementCount - 1)` to always have the last element 200 mm from the first independent of ElementCount

Jolbas:
> There is something missing at the end:
> 
> `create(<<placement>>; create(<<Vector>>; 1; 0; 0); create(<<rotation>>; 360 / ElementCount * ElementIndex; 0; 0); LCS_0.Placement.Base) * SourceObject.Placement`

Yes, seems to be good. And then also not mention the center argument. It became a mess that I couldn't get a grip of in combination with a SourceObject.Placement
`create(<<placement>>; create(<<Vector>>; 1; 0; 0); create(<<rotation>>; 360 / ElementCount * ElementIndex; 0; 0)) * SourceObject.Placement`

Zolko-123:
> Is it not confusing to both have X,Y,Z and also a direction? ... But we have to decide whether to default to XYZ displacement or direction+length.

well, that's decided then: we don't want XYZ but free directions. I need to do assemblies like this: 


vacuumvessel2

Jolbas:
That is of cause possible and we surely want an example of that in the documentation but the default state should show simple basic arrays, no? How do you use the direction property to accomplish that array?

Zolko-123:
look in the CircularArray code

Jolbas:
Your image shows some kind of spherical array and the CircularArray seems to just rotate around a single axis. Could be done by attaching the array to an axis. But that is not thought through and is maybe a bad idea.

Zolko-123:
the image shows a bunch of circular arrays rotating around a single axis (the holes for screws) in many 3D directions.

Jolbas:
So you want to be able to place screws in all those holes using one single ExpressionArray instead of attaching one array to each pipe center? That's probably possible but cumbersome.

Jolbas:
Another less practical spherical pattern example:
`create(<<placement>>; create(<<vector>>; 0; 0; 0); create(<<rotation>>; 100 * ElementIndex; 0; asin(2 * ElementIndex / (ElementCount - 1) - 1))) * .SourceObject.Placement`
heart sphere

Zolko-123:
I want to place an array of screws on each flange, meaning I want to place X circular arrays each along an arbitrary direction.

Concretely, I expect that the vacuum vessel contains datum axes along each of its flanges, and I use those datum axis for each circular array.

Zolko-123:
(Nice spherical array BTW)

Zolko-123:
Hello @Jolbas , I've updated the _ExpressionArray_ object in the file _Asm4_objects_. Can you please update and comment on the fields in there ? (WARNING: there are hidden properties).

I'd like now to create a new command _makeLinearArrayCmd_ that uses this _ExpressionArray_ object to make a linear array along the **Direction** line. It should therefore add a new property "StepSize" (a float probably) and use that.

PR #354, opened 2022-09-21
BOM with Fastners

leoheck:
This is an attempt to add fasteners to the BOM. 
It works, needs test.

### DIFF
--- a/FastenersLib.py
+++ b/FastenersLib.py
@@ -273,7 +273,7 @@ class cloneFastenersToAxesCmd():
                                     
             Gui.Selection.clearSelection()
             self.rootAssembly = Asm4.getAssembly()
-            if rootAssembly:
+            if self.rootAssembly:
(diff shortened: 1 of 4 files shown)


Zolko-123:
thank-you. I won't really be able to test it for some days though, so fingers crossed

leoheck:
That's fine. I will be testing this myself for a while. 

There is something that I still have to think about. For instance the Link to a Model is being inserted in the BOM with it's sub parts. This doesn't look right. It should be one or another, I guess. The later is better, I guess.

Also, the quantity is increased using the the label only and not the document name. In this case, the overall results (on my tests) look good but there are some cases where the document name is from different places and then this is a bit confusing.

Ah, there is also a lot of text to improve. Most of the things are hard to use because it is hard to understand. And the code is a bit unformated, it deserves a bit of love too.

Issue #358, opened 2022-09-24
Part creation changing Parts name


leoheck:
Hi @Zolko-123 is there any special reason for Asm4 changing the characters the user has put in the Part name?

Please, look the following behavior. I want to create this `name1_name2-name3.name4` Part

![image](https://user-images.githubusercontent.com/1277920/192098282-a66b7c51-9f33-4ee4-9c94-cf386ba55579.png)

And it ends up being created as `name1_name2_name3_name4`

![image](https://user-images.githubusercontent.com/1277920/192098298-312bf0aa-4aa5-4242-852f-c3c560b26a6c.png)

I would keep the user's choice unless there is a hard reason for that. But I don't see why it needs to be this way since the user can change the Part name after the creation of it. 

![image](https://user-images.githubusercontent.com/1277920/192098466-b8bdd779-a774-42dd-a306-b0af14929c15.png)

This is a bit confusing especially when creating a bill of materials. 

Is this a bug or something really important?

leoheck:
Ah, ok. The add `addObject` modifie the name to be an internal name, for some reason. 
Then it adds the same as Label.

To keep with what the user wants, it needs something like `newPart.Label = text` after using the `addObject`
I added it here in the line 80. 

![image](https://user-images.githubusercontent.com/1277920/192099142-64669834-30d1-4335-b0bd-ea08bbd6ec94.png)

Zolko-123:
good catch, thank-you.

PR #359, opened 2022-09-24
Set the label of a Part after its creation

leoheck:
Relates with https://github.com/Zolko-123/FreeCAD_Assembly4/issues/358


Zolko-123:
I didn't see your PR, so I did this already after your remark in the issues. Thanx

PR #363, opened 2022-09-29
Late improvements to the bill of materials maker.

leoheck:
Late improvements to the bill of materials maker.

- Adds `PartDesign::Body` objects
- Handles `Part::Feature` objects
- Skips (Asm4) Parts when there are other objects inside 
- Skips the Model file since it adds its parts 
- Skips hidden objects and links

Now BOM is more or less working the way I visualized it. 
However, I still have to check with bigger and deeper designs. 
Good, I am building one, so I will keep testing it.

### DIFF
--- a/InfoKeys.py
+++ b/InfoKeys.py
@@ -83,32 +83,38 @@ def infoDefault(self):
                     try:
                         sketch = pad.Profile[0]
                     except NameError :
-                        print('There is no Sketch on a Pad of the Part', part.FullName)
+                        # print('There is no Sketch on a Pad of the Part', part.FullName)
+                        pass
 
         try:
             LabelDoc(self, part, doc)
         except NameError:
-            print('LabelDoc: there is no Document on the Part ', part.FullName)
+            # print('LabelDoc: there is no Document on the Part ', part.FullName)
+            pass
 
         try:
             LabelPart(self, part)
         except NameError:
-            print('LabelPart: Part does not exist')
+            # print('LabelPart: Part does not exist')
+            pass
 
         try:
             PadLength(self, part, pad)
         except NameError:
-            print('PadLenght: there is no Pad in the Part ', part.FullName)
+            # print('PadLenght: there is no Pad in the Part ', part.FullName)
+            pass
 
         try:
             ShapeLength(self, part, sketch)
         except NameError:
-            print('ShapeLength: there is no Sketch in the Part ', part.FullName)
+            # print('ShapeLength: there is no Sketch in the Part ', part.FullName)
+            pass
 
         try:
             ShapeVolume(self, part, body)
-        except:
-            print('ShapeVolume: there is no Shape in the Part ', part.FullName)
+        except NameError:
+            # print('ShapeVolume: there is no Shape in the Part ', part.FullName)
+            pass
 
 
 
--- a/makeBomCmd.py
+++ b/makeBomCmd.py
@@ -72,7 +72,7 @@ class makeBOM:
         self.modelDoc = App.ActiveDocument
         try:
             self.model = self.modelDoc.Assembly
-            print("BOM of Assembly 4 Model")
+            print("BOM of the Assembly 4 Model")
         except:
             try:
                 self.model = self.modelDoc.Model
--- a/makeBomCmd.py
+++ b/makeBomCmd.py
@@ -72,7 +72,7 @@ class makeBOM:
         self.modelDoc = App.ActiveDocument
         try:
             self.model = self.modelDoc.Assembly
-            print("BOM for Assembly 4 Model")
+            print("BOM of Assembly 4 Model")
         except:
             try:
                 self.model = self.modelDoc.Model
@@ -88,7 +88,11 @@ class makeBOM:
         self.inSpreadsheet()
         self.BOM.setPlainText(self.Verbose)
 
-    def listParts(self, obj, level=0):
+    def indent(self, level, tag="|"):
+        spaces = (level + 1) * "  "
+        return "[{level}]{spaces}{tag} ".format(level=str(level), tag=tag, spaces=spaces)
+
+    def listParts(self, obj, level=0, parent=None, grandparent=None, asm4_obj=False):
 
         file = open(ConfUserFilejson, 'r')
         self.infoKeysUser = json.load(file).copy()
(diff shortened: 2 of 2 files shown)


Zolko-123:
Thanks. I can't test it right now so I merge it blindly. I still checked that the PR doesn't modify files outside of its perimeter.

hasecilu:
Hello, as you recommended I tried the `development` branch but got an error,
The first ~200 lines of log are parts being identified and then this error:
```
Traceback (most recent call last):
212   File "/home/uli/snap/freecad-realthunder/common/.FreeCAD/Mod/Assembly4/makeBomCmd.py", line 88, in Activated
213     self.inSpreadsheet()
214   File "/home/uli/snap/freecad-realthunder/common/.FreeCAD/Mod/Assembly4/makeBomCmd.py", line 328, in inSpreadsheet
215     data = sorted(data, key=lambda d: d['Doc_Label'])
216   File "/home/uli/snap/freecad-realthunder/common/.FreeCAD/Mod/Assembly4/makeBomCmd.py", line 328, in <lambda>
217     data = sorted(data, key=lambda d: d['Doc_Label'])
218 
219 ('Doc_Label',)
```

My design it's a little big, if you want you can test with it: https://gitlab.com/hasecilu/diy-pcb-cnc

leoheck:
@hasecilu thanks for testing it and also for sharing your design. It is awesome.
I changed the Column names. So this issue may be related to this change. 

This info is stored in this file here:
```
/home/$USER/.local/share/FreeCAD/Templates/Asm4_infoPartConf.json
```

As you can see, I could generate the BOM, and here is a preview of it.
image

This is the way "I think" you can fix/update this .json file.

Go to one of the parts you have, and then click this `Edit Part Information
image

And then, here, use the `Reset Fields`
image

`WARNING`: I am still not aware of the full usage of this button. So FIRST, backup this file to make sure it is not going to lose any important data, and since your repo is versioned, there is no problem on the design files:
```
/home/$USER/.local/share/FreeCAD/Templates/Asm4_infoPartConf.json
```

Off-topic, interesting choice of name here `ASSembly.FCStd` haha

leoheck:
I was comparing the generated screws with your `Fastners_BOM` looks like some are missing. Maybe related with the folders. I am not sure I am handling folders. I am going to improve this.

If you can run it, can you please, check if there is something missing, wrong or something that you are not expecting?

leoheck:
I got confused the first time, but it is working, actually.

hasecilu:
I was trying to make the json fix but the `Edit Part Information` gives me an error, **probably something in my installation is broken**. I'm using the latest snap of the Link Branch, I'll try later with a previous AppImage and also check with the FreeCAD 0.21 AppImage.
About the`Fasteners_BOM`, that file was created with the `Fasteners WB`

leoheck:
I am using this version here

```
OS: Ubuntu 22.04.1 LTS (ubuntu:GNOME/ubuntu)
Word size of FreeCAD: 64-bit
Version: 0.21.30454 (Git) AppImage
Build type: Release
Branch: master
Hash: 5df954690787dedbbf3c35f658f010f388474676
Python 3.10.6, Qt 5.15.4, Coin 4.0.0, Vtk 9.1.0, OCC 7.6.3
Locale: English/United States (en_US)
Installed mods: 
  * Defeaturing 1.2.0
  * A2plus 0.4.59c
  * Curves 0.5.8
  * fasteners 0.4.12
  * lattice2 1.0.0
  * Assembly4 developer branch <==============
  * kicadStepUpMod 10.15.1
  * Help 1.0.3
```

leoheck:
> About theFasteners_BOM, that file was created with the Fasteners WB, if I can remember correctly only the fastener used in the actual file are listed, so, in every sub-assembly I should make a fasteners-BOM.

Got it. The way it is working know is to generate the whole BOM where it was run... you can generate a BOM on each assembly. This would be good to check if things are working with less complexity too. 

The only issue is that parts are being grouped regarding the `Part_Label` but not considering the `Doc_Label`. 
This is good since we are going to have a single entry per each Part. 
This is bad because maybe different parts with the same name are going to be grouped and also, we cannot keep track on where each part is going to be installed since the Doc_Label is not uefull anymore..

I will have to discuss which approach is better. If it is better to keep as it is, or split parts considering the Doc_Label. 

Considering the Doc_label is good too because the user can group parts later. Now if it does not consider Doc_Label then groups cannot be reverted (only by looking the Spreadsheet) since we are losing this information.

leoheck:
Here I am comparing your Fastner_BOM with the ones gotten form the generated BOM file
image
Only lines 3 and 5 match. 

But your design has some Fastners that are kind of strange to me. Some are just Parts and are not Fatners. You have to investigate the generated BOM closely, to check if it is ok or not since I cannot say this easily.

image

hasecilu:
> I changed the Column names. So this issue may be related to this change.

Yes, it looks like this was the problem. I tried updating the file the way you described but the button didn't work, I ended up copying the `json` file manually and did the job.

> If you can run it, can you please, check if there is something missing, wrong or something that you are not expecting?

For the fasteners BOM I still get the same BOM with the Fasteners WB
|Type|	Qty|
|---|---|
|ISO4032 |Nut M4	|4|
|ISO4762| Screw M3x8	|4|
|ISO4762| Screw M4x8|	4|
|ISO4762 |Screw M5x10|	2|

>  Maybe related with the folders. I am not sure I am handling folders. 

I remember few months ago I made this change and the `Parts` that were inside a `Group` (folder) were adde to the BOM: https://github.com/howie-j/FreeCAD_Assembly4/commit/18d9483f8ddc34b5db081bda336e4de68e3278f9

> If you can run it, can you please, check if there is something missing, wrong or something that you are not expecting?

I think the links should not be skipped, but that's for later...

Here I included a spreadsheet showing the changes in the BOM from the `master` to the `development` branch 
[BOMs.ods](https://github.com/Zolko-123/FreeCAD_Assembly4/files/9690635/BOMs.ods)

leoheck:
I was not seeking the links in the bom, but when I was about to implement something to find the Folder, they started to appear in the BOM. Even with me reverting the code to the original version of the development branch. 

Can you try to delete de BOM file manually, and create it again. Just to make sure they are going to appear.

hasecilu:
> Can you try to delete de BOM file manually, and create it again. Just to make sure they are going to appear.

Yes, the whole BOM reappears.

leoheck:
If the links are included, then if you are Linking the Car#Model, you are going to see it in the BOM with its own Parts.
Like: 

- Car <===== This car is extra since Wheel + Door + Window are the car
- Wheel
- Door
- Window

I feel that it is better to keep in the BOM, only the final parts that you have to buy or build.

Now, the Assembled Part could be Added considering we need or have a Car ready for Assembly.
Then, we could have Car_BOM generated in the same file, too.
This solution would be interesting too. 

Actually, this BOM subject would be better if it had a dialog configuring what the user wanted... 
But I believe this is beyond my current ability.

leoheck:
> Yes, the whole BOM reappears.

Sorry, I deleted the previous message, since I realized this was not your problem.

hasecilu:
Maybe there is a misunderstanding with the `Links`, the `Links` are basically the copies/clones of a `Part` and not their components.
For example in my project in the `Bed_top` file generate the BOM and you'll note that the quantities of the pieces don't correspond with the ones that are in the assembly and it's because the `Element count` property of the Link is ignored.

leoheck:
Alright, using your `Bed_top`, I am seeing that indeed BOM is missing parts inside folders. I was fixing this yesterday, but then I stoped because I tought it was not a problem. It actually is, and I will tacle that right now. 

The quantities are wrong, becasue the Folders are being ignored. 

Let me check how can I improve this!

hasecilu:
> The quantities are wrong, becasue the Folders are being ignored.

That's the first half of the problem, the other is the `Element Count` property, check [this](https://wiki.freecadweb.org/Std_LinkMake#Data). In my case 9 Al profile are expected in the BOM but get 1.

Screenshot from 2022-10-01 15-07-44

leoheck:
This object does not seem to be an ASM4 Object. 
Maybe it was created with a different (old) version of it.
If you click on this object, and then "Edit Placement of a Part" you are going to see this.
Let me check if updatating this object changes something.
image


Since I have the Folders working, I am currently trying to undestand why I am seeing only one SC8UU object.
But this one is also a problem since it is composed with other objects and those are not in the bom.
Now SC8UU is not a single thing. It is composed with multipls things. So each thing, once the Folders are included, 
image
image

leoheck:
If I click "Go to the Link" in this `2020x370mm` object, nothing happens.

leoheck:
This is also not an Asm4 link

leoheck:
Inside the Model is allowed only Links, Fastners (and Folders, on my code here)

leoheck:
I mean, I am not sure if this is the something Asm4 is expecting. It looks like you are using App:Part to group App:Links

leoheck:
Ah, I found the issue, 
Help me to check, please.

image

hasecilu:
>  But this one is also a problem since it is composed with other objects and those are not in the bom.

And it should be like that because SC8UU is a `Part` not an assembly. That the reason the Center Metal Case objects are `Bodies`. You buy SC8UU in the store not it's parts.
As a counter part if you check the main BOM you will notices several entries of `NEMA_17_witth_connector` in the `Doc_label` because the model of the motor was created with several `Parts` inside another `Part` and this is a bad practice (or at least bas for the BOM) because in the BOM you don't buy the parts of the motor, you buy the motor!

Zolko-123:
> It looks like you are using App:Part to group App:Links

This is a very strange (and bad) idea. If anything use a Group, or else a real sub-assembly (in an external file).

leoheck:
Some of the issues are becuase maybe your design is using old versions of Asm4 (this is just a guess).
Yeah, motor is an Assembly of objects that is not An Assembly Part. I was looking for a way to easily identify this. Since I dont want to modify Asm4 files, I am considering the Part that is right inside the Link is an ASM4 part (alternatively it should have an LCS inside, but this is not the best case because LCS can be everywhere)

In the case of this Motor, I would do this

Create a ASM4 Part named NEMA_Motor, and inside of it a Part named NAMA_Motor (inside it will have its parts)
Now, when you link it, you are going to have something like this

Link Name > (ASM4) NEMA_Motor > NEMA_Motor > (A bunch of parts)
In this case, the thing the is going to be in the bom is the in the NEMA_Motor since it does not have any Link direct to it.

leoheck:
Your NEMA_17_witth_connector, is made in a bad way from my point of view to have a easy to use Part.
Becuase, if you do this, you are telling me the internal parts should be included in the BOM.

image

If you dont want to have parts inside the motor listed, then you have to so something like this
Beucase you are using a Assembly (that is not Asm4) to build it. 
image

leoheck:
In other words. If you have an Asm4 Part or Body (this is indicated by having a link to this object, since there is not any property inside of them that tells me they are Asm4 objects) you want to list the nested objects in the BOM.

If you dont want to list subojects in the BOM, they have to be placed inside another Part. This will hide them, because this new part is not an Assembly4 part. And was not created using LCSs.

Zolko-123:
I think that the BOM shouldn't go inside Parts (yellow icon) and Bodies (blue icon) and links to them (they are final objects)

When a link is to a Part and it's an Asm4 Assembly then we can/should go look inside it. But then there should be a limit of how deep to go because for large assemblies this could be HUGE. 

Thinking of this, there should be a function in Asm4_libs called `isAsm4Model()` that checks whether an object (in this case the LinkedObject) is an Asm4 Model of Type Assembly.

leoheck:
Other thing I recomend is to use the something that transforms this Motor in a single Part, so you can place LCS and keep them under the Asm4 object where they can be linked.

leoheck:
> I think that the BOM shouldn't go inside Parts (yellow icon) and Bodies (blue icon) and links to them (they are final objects)

Me too. And this is what it is doing.
I would like to create an image explaining that, but I dont have it yet, since it was not wasy to create a good one that covers what I am thinking, but I will try.

> When a link is to a Part and it's an Asm4 Assembly then we can/should go look inside it. But then there should be a limit of how deep to go because for large assemblies this could be HUGE.

It limits to only one Level. But sub assemblies can go deep forever. Which is fine. 

> Thinking of this, there should be a function in Asm4_libs called isAsm4Model() that checks whether an object (in this case the LinkedObject) is an Asm4 Model of Type Assembly.

I would appreciate this, Indeed.

leoheck:
This is my Summary, let me know if it is understandable.

These are the things I am handling in the BOM

```
Model ================================================== Does not go in the BOM
  App:Link --------------------------------------------- Does not go in the BOM
    App::Part ------------------------------------------ Does not go in the BOM **
    PartDesign::Body ----------------------------------- Goes in the BOM
    Part::Feature -------------------------------------- Goes in the BOM
  Part::FeaturePython' > "FastenersCmd" ---------------- Goes in the BOM
  App::DocumentObjectGroup (have to commit this yet) --- Does not go in the BOM

** Only App::Part, Part::Feature and PartDesign::Body inside of the linked App::Part (1st level only) go in the BOM
```

Zolko-123:
> > Thinking of this, there should be a function in Asm4_libs called isAsm4Model() that checks whether an object (in this case the LinkedObject) is an Asm4 Model of Type Assembly.
> 
> I would appreciate this, Indeed.

done (in the development branch)

leoheck:
Regarding this, the issues I saw and I am fixing right now are.
- BOM does not handle Folders inside de Model (solved already)
- it was not checking if PartDesign::Body was already added to increase quantity of this objects. (done already)
- Now, use this new isAsm4Model() instead of what I was using. (doing it)

Zolko-123:
How about:

```
Model ================================================== Does not go in the BOM
  App:Link --------------------------------------------- Does not go in the BOM
    App::Part ------------------------------------------ Goes in the BOM
    PartDesign::Body ----------------------------------- Goes in the BOM
    Part::Feature -------------------------------------- Goes in the BOM
    Asm4::Model ---------------------------------------- User decides whether to go in the BOM **
  Part::FeaturePython' > "FastenersCmd" ---------------- Goes in the BOM
  App::DocumentObjectGroup (have to commit this yet) --- Does not go in the BOM
```
** there is an option to say if the BOM should stop at assembly level (sub-assembly included as a whole in the BOM) or look inside sub-assemblies (BOM of sub-assembly included) . But I don't know how difficult it is to do that

leoheck:
> ** there is an option to say if the BOM should stop at assembly level (sub-assembly included as a whole in the BOM) or look inside sub-assemblies (BOM of sub-assembly included) . But I don't know how difficult it is to do that

I think the hard part for me is to implement a window with the option.
But if you are saying that we can have 2 buttons in the toolbar (or maybe a list). 
Where the first follows the subassemblies and the second just lists objects in the current file. I think won't be hard since I can copy one of the existing buttons and can do this to my next PR with the fixes.

hasecilu:
Yes, most of the assemblies were made months ago.
Regarding the motor, it could be fixed making all the bodies child of the main Part and removing the remaining (and empty Parts), as the LCS are child of the main Part there is no chance to break.
Note that this file is not an Assembly4 file, just a regular Part.
Alternatively it can just be a step file inside a Part container.
Screenshot from 2022-10-01 16-30-11

hasecilu:
> If I click "Go to the Link" in this `2020x370mm` object, nothing happens.

If should open the V profile file, the whole project was created with the Link Branch version from realthunder, maybe because of that you can not open it in 0.21v

hasecilu:
>  Help me to check, please.
In main ->  SC8UU &  T8 antibacklash are ok, 2020x370mm should be 9, spacers: 20
In bed_top -> fasteners: 18, 16 & 4, sliding nuts: 18

hasecilu:
The reason why with SC8UU is correct in the BOM is because I manually added a Part with the `Asm4_insertLink` function 4 times in 3 different files, so it's counted 12 times and that appears in BOM.
For `2020x370mm` and Fasteners I made `Asm4_insertLink` and used the `Element count` property of the Link, so just one LCS union and several offsets.

leoheck:
I think I have somehting. I am going to make a PR, so we can discuss there. Cool?

leoheck:
Here, https://github.com/Zolko-123/FreeCAD_Assembly4/pull/366

PR #366, opened 2022-10-01
[WIP] BOM

leoheck:
Add support for Folders inside the Model
Fix the amount of PartDesign objects that were not being increased 
Add extra command to list only local parts (skipping subassemblies)

TODO: Still have to use this new isAsm4Model() function.


leoheck:
@Zolko-123 I am reusing the same BOM class to create 2 buttons. Each one runs the code and creates a BOM (with the spreadsheets called BOM and BOM_Local) however both files are generated and selected in the end no mater which button is being clicked. Do you have an idea of how I can fix this?

hasecilu:
Is the actual proposal like this:
```
Model ================================================== Does not go in the BOM
  App:Link --------------------------------------------- Does not go in the BOM
    App::Part ------------------------------------------ Does not go in the BOM **
    PartDesign::Body ----------------------------------- Goes in the BOM
    Part::Feature -------------------------------------- Goes in the BOM
  Part::FeaturePython' > "FastenersCmd" ---------------- Goes in the BOM
  App::DocumentObjectGroup (have to commit this yet) --- Does not go in the BOM
```
** Only the 1st level inside of the linked App::Part go in the BOM (App::Part, Part::Feature and PartDesign::Body) 

For example in this sub-assembly I use 2 KP08 units but as you can see in the BOM it appear each one of its components, so I don't think Bodies or Features should go in BOM. Otherwise you know how can I "isolate"all those bodies?

My project has a `testing` branch if you wanna see the changes.
Body_problem

leoheck:
Sorry, the current branch is not quite working. I am trying my best to improve it, but some things are getting weird. 

I see what you are saying, but I had a special and important case to allow that. But just in the first level inside of the Part. But I cannot remember my point. I have to think.

leoheck:
This is promising, could you check when you have time?

This is the Bed_top
My own files continue working. And some of your files are better.
Your main ASSembly still has some details to take care of. But I believe it is from your side upgrading the design.
However, if you see something important, let me know, please.
image

leoheck:
This view is when we are including sub-assemblies.
image

Zolko-123:
I'll make later a makeBOM window, for now this looks quite good

leoheck:
Thanks, this time behaves better indeed.

hasecilu:
> This view is when we are including sub-assemblies.

Yes, I have just tested all my assemblies and the result is very consistent and according what it's expected.
I have updated my `testing` branch with latest changes

PR #368, opened 2022-10-02
Improve BOM part 3

leoheck:
It Fixes the greedy walk for some specific cases.
Improves console messages with better information
Present parts in the same order they appear, making it easier to verify if the BOM is right.

Now my designs that were made recently with 0.21 and Asm4 work nicely. And also, the design of @hasecilu that is big, complex but quite right. 

This is a late fix. Sorry for too many PRs @Zolko-123

### DIFF
--- a/makeBomCmd.py
+++ b/makeBomCmd.py
@@ -142,7 +142,7 @@ class makeBOM:
                     self.Verbose += "- linked: {linked_obj}\n".format(linked_obj=obj.LinkedObject.Name)
                 except:
                     self.Verbose += "- linked: {linked_obj}\n".format(linked_obj=obj.Name)
-                self.Verbose += '- not included in the BOM\n\n'
+                self.Verbose += '- not included\n\n'
 
                 self.listParts(obj.LinkedObject, level + 1, parent=obj)
 
@@ -185,7 +185,7 @@ class makeBOM:
                         print("ASM4> {level}| {qtd}x | {obj_typeid} | {obj_name} | {obj_label}".format(level=self.indent(level, tag=" "), obj_label=obj_label, obj_name=obj.FullName, obj_typeid=obj.TypeId, qtd=qtd))
 
                         self.Verbose += "> {level} | {type}: {label}, {fullname}\n".format(level=obj_label, type="ASM4_PART", label=obj_label, fullname=obj.FullName)
-                        self.Verbose += "- object already added\n\n"
+                        self.Verbose += "- object already added (" + str(qtd) + ")\n\n"
                         self.PartsList[obj_label]['Qty.'] = qtd
 
                 else:
@@ -193,6 +193,7 @@ class makeBOM:
                     print("ASM4> {level}| 1x | {obj_typeid} | {obj_name} | {obj_label}".format(level=self.indent(level, tag=" "), obj_label=obj_label, obj_name=obj.FullName, obj_typeid=obj.TypeId))
 
                     self.Verbose += "> {level} | {type}: {label}, {fullname}\n".format(level=obj_label, type="ASM4_PART", label=obj_label, fullname=obj.FullName)
+                    self.Verbose += "- adding object (1)\n"
 
                     self.PartsList[obj_label] = dict()
                     for prop in self.infoKeysUser:
@@ -200,9 +201,11 @@ class makeBOM:
                         if self.infoKeysUser.get(prop).get('active'):
                             try: # to get partInfo
                                 getattr(obj, self.infoKeysUser.get(prop).get('userData'))
+                                info = "(predefined)"
                             except AttributeError:
                                 crea(self,obj)
                                 fill(obj)
+                                info = "(extracted)"
 
                             if self.infoKeysUser.get(prop).get('visible'):
                                 data = getattr(obj, self.infoKeysUser.get(prop).get('userData'))
@@ -216,7 +219,7 @@ class makeBOM:
                                 data = obj_label
 
                             if data != "-":
-                                self.Verbose += "- " + prop + ': ' + data + '\n'
+                                self.Verbose += "- " + prop + ": " + data + " " + info + "\n"
 
                             self.PartsList[obj_label][self.infoKeysUser.get(prop).get('userData')] = data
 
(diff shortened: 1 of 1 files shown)


leoheck:
Ah found something else.
Links of Fasteners.
This may be a bug, but I think it is a unsupported feature.

@hasecilu how did you make those links?

@Zolko-123, does Asm4 should accept having links to Fasteners?
If so, I have to add this part to be treated in the BOM

image

hasecilu:
I made an update minutes ago in my `testing` branch, (don't include this PR).

I included the Fasterners_BOM created by the Fasteners WB and it includes the correct number of Fasteners, in this script (https://github.com/shaise/FreeCAD_FastenersWB/blob/master/FastenerBase.py) the `GetTotalObjectRepeats(obj)` seems to be the function that count elements and it's able to count `Links` like the ones I used a lot, maybe it could be taken as reference. I think the magic line is the 315

leoheck:
cool, I will test. but the messages here are still good and valid. 
i am going to upgrade the pr accordingly.
but I am going to have lunch now, so I come back right after

hasecilu:
> @hasecilu how did you make those links?

I made the links this way: https://youtu.be/tltsTgfC3U8?t=51
The same for Link in fasteners and other models like Sliding nuts or some Al profiles

leoheck:
It looks like it has to be supported since Fasteners are. I will check how to do it.

leoheck:
Almost there. It works for the ASSembly.FCStd file (fastener links inside a folder)
When the fastener link is right under the Model, the count is 1 less. =/

hasecilu:
I have just tested the code and can confirm it, it's odd

leoheck:
I could finish!

leoheck:
I am about to commit.

leoheck:
Alright, now it looks like it is working for all the scenarios I am aware of.

hasecilu:
The last fix is working.
I detected another scenario: when you use the same type of fastener without using the `Link` function both appear in the BOM but in different rows.

I like to use the `Link` function because the new fasteners inherit the properties of the main one like type, length, material, etc. It's totally valid don't using `Links` but you need to change manually for every copy.

Same_fast

leoheck:
I tried that, it was working for me. I think you are doing some black magic haha. Well, if this SW_X file is in your repo I will try to check this file myself to see if it is different here.

leoheck:
I added the screw and nuts, and here, they are right.
image
image

leoheck:
What is the version of your Freecad again?

leoheck:
Ah, I updated the repo, and your file is there with your screws, got the same behavior.

leoheck:
It is a bit odd the way you add the screws. I am using the ones in the Asm4 panel.

leoheck:
Console log, the last column (obj.Label) is different from your file and mine. 
I could use the obj.Name to check if they are the same. But this is not good =/ since this cannot be changed by the user

# hasecilu
```
21:33:55  ASM4> [1]    | App::Link | SW_X#LimitSW_CNC_1 | LimitSW_CNC 1
21:33:55  ASM4> [2]        | 1x | App::Part | Limit_SW_PCB#Part001 | LimitSW_CNC 1
21:33:55  ASM4> [1]      | 1x | Part::FeaturePython | SW_X#Screw | M3x14-Screw
21:33:55  ASM4> [1]      | 1x | Part::FeaturePython | SW_X#Screw001 | M3x14-Screw001
21:33:55  ASM4> [1]      | 1x | Part::FeaturePython | SW_X#Nut | M3-Nut
21:33:55  ASM4> [1]      | 1x | Part::FeaturePython | SW_X#Nut001 | M3-Nut001
```
image


# MINE
```
21:36:21  ASM4> [1]    | App::Link | SW_X#LimitSW_CNC_1 | LimitSW_CNC 1
21:36:21  ASM4> [2]        | 1x | App::Part | Limit_SW_PCB#Part001 | LimitSW_CNC 1
21:36:21  ASM4> [1]      | 1x | Part::FeaturePython | SW_X#Screw | M3x8-Screw
21:36:21  ASM4> [1]      | 2x | Part::FeaturePython | SW_X#Screw001 | M3x8-Screw
21:36:21  ASM4> [1]      | 1x | Part::FeaturePython | SW_X#Nut | M3-Nut
21:36:21  ASM4> [1]      | 2x | Part::FeaturePython | SW_X#Nut001 | M3-Nut
```

image

leoheck:
When I update something, the fasteners on your design are normalized to have the same name (I mean, label)
image

hasecilu:
> What is the version of your Freecad again?

snap of FreeCAD Link Branch edge branch, it contains the latest changes of reathunder, I have tried to open it in 0.21 but the ASSembly gets scrambled.

hasecilu:
> It is a bit odd the way you add the screws. I am using the ones in the Asm4 panel.

I used the Fastener button included in Assembly4

leoheck:
And the version of the Fastener WB?

leoheck:
A fix that I am making is to remove the digits from the end of the Label... considering the Fastners did not normalize the names yet. But I would like to know how to normalize fasteners names.

hasecilu:
I changed of computer, but is one of the latest versions, less than a week at most

leoheck:
This last commit will solve this particular issue with Fastners.

leoheck:
Could you check?

leoheck:
I bet we can find more issues, but they are definitely getting harder to find.

leoheck:
Sure, take your time.

hasecilu:
In `Edit > Preferences >  General > Document > Document objects `, do you have checked "Allow duplicate objects labels in one document"? I don't and thats why I think in my file it gets the 001 suffix.

leoheck:
You are right, I have this checked. But I thought the renaming process was a feature of the Fasteners.  Actually, you are right. But the last fix should work for you too.

hasecilu:
Yeah, the latest commit fixed it!

leoheck:
Good! Thanks, @Zolko-123 this is ready to be merged.

hasecilu:
Now we can wait to hear users with different workflows if the new BOM works for them, having 2 separate BOM options is really handy. Thanks for your work!

leoheck:
Totally. If you know someone, just ask them to test it, please.

Zolko-123:
Thank-you. Unfortunately I have no time to test in these days, sorry

PR #373, opened 2022-10-15
Rework of Expression Array

Jolbas:
More info in [this post](https://forum.freecad.org/viewtopic.php?p=633241#p633241)


Zolko-123:
Please don't mix white-space clean-ups with real code. Also there is a conflict with _makeArrayCmd_

Jolbas:
Ok. Sorry. I fix it. I had some kind of cleanup tool installed for an other projekt and didn't notice the change until it was too late.

PR #375, opened 2022-10-16
Stabilize auto-numbering behavior

abetis:
More stable solution for the auto-numbering.
Asm4.nextInstance will always add "/001", "/002" and so on to the name.
User can still specify his own numbering in the same concept if needed, but better leave the logic do it's own work automatically.

Previous implementation stripped numbers from the name even when not added anything, which cause a mess for parts names that include "_number" at the end.
Since FreeCAD naming converts many characters that it doesn't like to undescore without any notice, meaning many names might end up into that category, including those that have dashes "-" and other characters as well.

### DIFF
--- a/Asm4_libs.py
+++ b/Asm4_libs.py
@@ -344,20 +344,32 @@ def getDependenciesList( CompleteSelection ):
     |           get the next instance's name        |
     +-----------------------------------------------+
 """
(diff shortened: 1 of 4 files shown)


Zolko-123:
please don't be silly

PR #376, opened 2022-10-16
Select the proper entry in the parent list when a part was clicked during link addition

abetis:
Helps alot when the LCS is not visible, or a part have many LCSs and user wants to try some of them from the list to see what matches best

### DIFF
--- a/placeLinkUI.py
+++ b/placeLinkUI.py
@@ -443,6 +443,7 @@ class placeLinkUI():
         # selected object
         selObj = Gui.Selection.getSelection()[0]
         selPath = Asm4.getSelectionPath(doc, obj, sub)
+
         # check that a datum object is selected
         if selObj.TypeId in Asm4.datumTypes and len(selPath) > 2:
             selLink = self.activeDoc.getObject(selPath[2])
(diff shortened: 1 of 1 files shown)


Zolko-123:
I don't understand what this does

abetis:
When adding a link to a model, you get the dialog where you have to select the parent model where you want to connect the new model, and the LCSs on both new and the parent models that should be attached.
You can also click on the 3D view on the parent LCS to populate the parent selection dropdown and parent LCS list automatically (that code exist and it's very helpful).

My addition allows to click on a "body" of a model in the 3D view to populate only the parent model dropdown. The user will have to select the LCS from the list as an additional step.
This is very helpful when you have many instances of the same model in the assembly and the parent drop down list is very long with similar names.
In my case I also have many times situations where the parent LCSs are hidden on the 3D view (either I hide the LCSs before, or the LCS where the new part should be connected is under another parts and not visible - happens a lot for imported models with hundreds of connection points, where I define only few LCSs and move the attachment point to the needed location during the addition phase).

That way the workflow is more intuitive with the 3D view clicks.

Zolko-123:
OK, yes, that's good, thank-you

PR #379, opened 2022-10-16
Rework of ExpressionArray

Jolbas:
More info in [this post](https://forum.freecadweb.org/viewtopic.php?p=633241#p633241)


Jolbas:
I changed the way to keep track of the axis. Previously the name of the axis or a sub element axis from a LCS was saved in a parameter. Now it's an expression containing a link to the axis and a modifier if a LCS subelement is selected. This is because I couldn't find a way to fetch any subelements in expression.
I'm not sure this change was good. From the beginning I planned to have the axis calculation visible in the ElementPlacement expression but I then realized this could remain in the python script along with the `... * SourceObject.Placement.` This way the expressions in ElementPlacement doesn't get overloaded and you can set expressions on each subparameter (X, Y, Z, Angle and Axis). It's also possible to set expressions on euler angles by opening the placement in the task panel.
It was probably easier for the user to change to another axis by just changing the name. So maybe revert to same axis property as in the old CircularArray?

Jolbas:
A third option is to have a proper link property for the axis object in combination with an enumeration property where you can choose X,Y or Z. Both are initially set correctly according to selection on creation.

Zolko-123:
Hum ... I'm not sure I like this new PR. My idea was to have a generic ExpressionArray (also as object in Asm4_objects) and also LinearArray and CircularArray objects which are derivatives of the ExpressionArray. These Linear and Circular arrays can then have additional properties as needed.  

What do you think ? As a first step, replace the current CircularArray by a new function based on the ExpressionArray. Can you do that ?

Jolbas:
I kept the old CircularArray to not break documents that uses them but it is inactivated so you can't create new ones. An alterantive would be to convert older CirvularArray instances to the new type but it's difficult to catch all edge cases.  Therefore  the new class that derivative from ExpressionArray is called PolarArray in the code. There may be a better solution to that.

Zolko-123:
what I mean is that the general _ExpressionArray_ class should not have these properties:

```
obj.addProperty('App::PropertyDistance',  'LinearStep',       'Array',
            'Distance between elements along Axis')
obj.addProperty('App::PropertyAngle',     'IntervalAngle',    'Array',
            'Default is an expression to spread elements equal over the full angle')
```

they should only be added if really used. An _ExpressionArray_ should still have a _Direction_ property, because most arrays will need such a reference. A _Direction_ should be any object having a `Placement` property ... may-be adding X/Y/Z as an option

OK, I agree for _CircularArray_. Can you then make a _LinearArray_ object based on _ExpressionArray_ ?

Zolko-123:
in other words: ideally, I'd like to have as `Asm4_object` only the _ExpressionArray_, and create circular and linear arrays in the `makeArrayCmd` command. The icon is adjusted by setting the _ArrayType_ at creation time. Do you think that's possible ?

Jolbas:
That seems possible and I give it a try within the next few days.

Jolbas:
I can't make the ShowElement property work very well. The child elements is shown but then the placements does not update which I think removes the main point of the ExpressionArray. Can I just make it immutable and hidden or do we miss any important feature by doing that?

Zolko-123:
yeah, make a try, whatever seems best, and we'll see then. Please tell me when you think I can merge the PR.

Jolbas:
Updated the tooltips too. And made a better icon for linear array. Removed all initial expressons on the plane ExpressionArray. There is also a selector to switch between XYZ of axis object. It selects the right one if a subaxis on a LCS is selected like the previos circular array. I think it's ready now.

Zolko-123:
Thanks, merged. I will review and test this week-end and hopefully release it with the next version of the WB.

Nice icon !

Jolbas:
I found an issue with expressions that made a mess of one of my ExpressionArrays. Se [forum thread](https://forum.freecad.org/viewtopic.php?f=8&t=73898). A workaround is to double the recompute before copying the placement to the child element. I think it could be a good idea to add the workaround until issue is fixed

Workaround at line 465t in Asm4_objects.py with context:
```
for i in range(obj.Count):
            obj.Index = i
            obj.recompute()
            # Double recompute, workaround for this issue
            # https://forum.freecad.org/viewtopic.php?f=8&t=73898
            obj.recompute()
            plaElmt = p0 * obj.ElementPlacement * p1
            plaList.append(plaElmt)

```

Jolbas:
Ignore the double recalculate workaround suggestion. It doesn't guarantee correct result anyway.

Zolko-123:
I released a new version (0.12.5) with this, and it's really great, thank-you. 

There could be a minor remark: when selecting an edge of an object, one would want that the direction of that edge is used, and not the placement of the object itself (think a line of a sketch). For example, we could set an "_AxisObject_" (which would be the sketch, currently called _Axis_) and an "_AxisDirection_" which could be X, Y or Z, but could also be "Edge5". So _AxisDirection_ would be a string (replacing the enumeration  _AxisXYZ_), and if its X, Y, Z then apply that, and if it's another string then look for the corresponding edge shape (if it's not found then apply "Z")

I have code in Asm4_Measure that does this already, in `getDir( self, shape )`. The `execute()` call of the _ExpressionArray_ would need to be updated in _Asm4_objects_ of course, but it seems doable.

What do you think ?

Jolbas:
Thanks.

I did have a similar thought of being able to select an edge as direction but I didn't immediately found a way to do it so the thought got lost. I think it's a good idea and I can give it a try. With your code in Asm4_Measure I have a good starting point.

Jolbas:
https://github.com/Zolko-123/FreeCAD_Assembly4/pull/389

PR #388, opened 2022-12-16
Fixed bug when editing part placement (link branch)

neuoy:
When using the link branch of FreeCAD ( https://github.com/realthunder/FreeCAD/releases/tag/2022.11.27-edge ), there is a bug when trying to modify a part placement: the UI is not filled with existing values for "Linked part", "Parent part", and corresponding LCS.

This appears to be because in link branch daily, the end of the string is different: "^ (-1)" instead of "^ -1"

A simple test allows to succeed in both situations (even though to be honest I don't know what this string means in the first place or why it's different in the link branch)


Zolko-123:
thank-you. 

(please make PRs against the development branch in the future)

Issue #387, opened 2022-12-15
Feature request: add constraints to more than two LCSes


QiangF:
I believe assembly 4 is the future. Using LCS simplifies everything. We really need not only constraints other fix distances or angles between two LCSes, but also constraint to three or more LCSes, just some simple constraints like coincident of LCS origines, parallel of LCS axes, angle of LCS axis to LCS plane. And Asm4 should have its own solver instead of relying on extra sketches. I believe the solver will be much simpler than assembly 3. There are many use cases https://forum.freecadweb.org/viewtopic.php?style=1&f=20&t=74456&p=647579#p647579 

Perhaps this is already planed. Zolko wrote on the forum:

Later, it should be possible to define degrees of freedom in the LCS and develop constraints to solve them (for example you can leave the rotation around an axis free, or a translation...). The number of degrees of freedom would be limited to those expressly authorised, thus the solver much simpler than having to solve all 6 degrees of freedom for all parts. This can be implemented at a later stage to provide a full assembly functionality that wouldn't lag the leading CAD systems. 

What needs to be done to make this happen?

Zolko-123:
My plan was to add the A2+ solver to Asm4. I already have created a branch for that : https://github.com/Zolko-123/FreeCAD_Assembly4/tree/solver

but I haven't have time to do it yet. If you have some available time, we could discuss this

PR #389, opened 2022-12-30
Features for ExpressionArray

Jolbas:
Added support for lines and circles as Axes. Added support for Scaling which led to a Mirror Array replacing mirrorPart. An evaluation order solver for expressions because the builtin resolver couldn't handle sub-propertys of a placement property. A limitation that hindered a lot of creative and simpler solutions. Side effect: Twice as fast. Shortened property names for them to fit better in expressions and Property view. Using a PropertyLinkSub as Axis which can hold subelements property, As a consequence, removed AxisXYZ


Jolbas:
I found a bug in the evaluation order function and am trying to fix it now

Zolko-123:
thanx

Zolko-123:
the mirror feature is also very nice ! The capability to chose an LCS axis for the symmetry plane is fantastic !!!

Jolbas:
Yes, that turned out to be useful. With this PR I am enough satisfied with the ExpressionArray to start encourage others to use it. I have for example no more needs renaming and rearranging parameters that could break already made documents. What do you think? Something that bother you in its structure?

Zolko-123:
for now I've not seen any showstopper, it's a great improvement overall, thank-you.

If you have some available time and motivation, the next BIG THING would be the inclusion of the A2+ solver with Asm4. See #387

Jolbas:
Ok. I will look at it and see if I can find motivation. If so, I will eventually find the time too.

PR #391, opened 2023-01-11
Match positive directions on LCS axis too

Jolbas:
When axis is X or Y on LCS. Match positive direction for remaining axis with Placer.Base positive directions

### DIFF
--- a/Asm4_libs.py
+++ b/Asm4_libs.py
@@ -552,12 +552,25 @@ class dropDownCmd:
 
 """
     +-----------------------------------------------+
+    |         vector constants                      |
+    +-----------------------------------------------+
+"""
+VEC_0 = App.Vector(0, 0, 0)
+VEC_X = App.Vector(1, 0, 0)
+VEC_Y = App.Vector(0, 1, 0)
+VEC_Z = App.Vector(0, 0, 1)
+VEC_T = App.Vector(1, 1, 1)
+
+
+
+"""
+    +-----------------------------------------------+
     |         the 3 base rotation Placements        |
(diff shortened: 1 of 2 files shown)


Jolbas:
The only reason is that they are constants and that way I avoid recreate them on every function call. Not that this is a very expensive calculation. Maybe a good idea to prefix them with underscore.

Zolko-123:
OK. But then, what about putting them in the file [Asm4_libs](https://github.com/Zolko-123/FreeCAD_Assembly4/blob/master/Asm4_libs.py#L554) as other global static parameters ?

Jolbas:
Absolutely. I fix it.
I noticed when looking as [Asm4_measure](https://github.com/Zolko-123/FreeCAD_Assembly4/blob/master/Asm4_Measure.py#L1043).py the methods isVector, isCircle, isLine, isSegment and isFlatFace are also present in [Asm4_libs](https://github.com/Zolko-123/FreeCAD_Assembly4/blob/master/Asm4_libs.py#L554) but as functions. How about using the Asm4_lib functions instead because they don't use the self argument anyway? I can fix in same PR if there are no reasons behind.
Then if you think the mirror command works fine as derived from ExpressionArray I can remove the entire mirrorPartCmd.py

Jolbas:
I make a separate cleanup PR in that case. Won't add more to this one.

Zolko-123:
Thanx. I might re-order and/or rename the constants to a more consistent naming.

Jolbas:
Sure. I didn't have a good name for the (1,1,1) vector.

PR #396, opened 2023-02-06
Support for ShowElement in ExpressionArray

Jolbas:
Added support for ShowElement to behave similar to ordinary Link Arrays and removed workaround to hide  ShowElement


Zolko-123:
couldn't check right now, but seems OK

PR #397, opened 2023-02-07
Some updates to documentation

Jolbas:
Also redirected the help menu item to open the instructions on github instead of showing "Lorem ipsum"

### DIFF
--- a/HelpCmd.py
+++ b/HelpCmd.py
@@ -7,16 +7,8 @@
 # HelpCmd.py
 
 
-import os
-
-from PySide import QtGui, QtCore
-import FreeCADGui as Gui
-import FreeCAD as App
-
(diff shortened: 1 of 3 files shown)


Zolko-123:
why do you want to remove `placeLinkCmd()` ?

Jolbas:
ops. I didn't want that.

Zolko-123:
Good idea to take in the instructions from the repo. I'll check how that works. Thanx

Jolbas:
> Good idea to take in the instructions from the repo. I'll check how that works. 

It won't work at all because I removed too many imports. I thought I tested it bu obviously I didn't. Fixed in https://github.com/Zolko-123/FreeCAD_Assembly4/pull/398

PR #398, opened 2023-02-09
Fix Asm4Help and rinse TECHMANUAL

Jolbas:
Techmanual contained a lot of information already present in INSTRUCTIONS.md which made it harder to maintain.
Fixed that to many imports was removed from HelpCmd.py in commit https://github.com/Zolko-123/FreeCAD_Assembly4/pull/397/commits/82b98f3e990cf147772ccc967da32aa0ae4b68c3


Jolbas:
I should have made two PR instead. Sorry.

PR #404, opened 2023-02-24
Load asm4 even if anytree is missing

leoheck:


Zolko-123:
yes, something like that, but I would be even more brutal: if anytree is not available, don't do any of this at all. Like:

```
has_anytree = False
try:
    from anytree import Node, RenderTree
    has_anytree = True
except ImportError:
    FCC.PrintWarning("WARNING : Pylib dependency missing = anytree, exportFiles is not available\n")
    pass

if has_anytree:
    ...
else:
    ...
```
and define dummy commands to prevent errors.

Zolko-123:
I committed a change to make the dependecy optional. At least now Asm4 starts again, please update as you see fit

Zolko-123:
This conflicts now with my workaround. Please update your local repo and improve this solution

leoheck:
What should I do? Don't you want me to update my repo?
Why did you close it?

leoheck:
The latest changes do not work when you do have the anytree. 

I am using the latest dev
image

I have the lib installed
image

And I am seeing this
image

However, I am updating a fix in a new PR.

leoheck:
Here it is https://github.com/Zolko-123/FreeCAD_Assembly4/pull/405

leoheck:
Thanks for your help.

PR #406, opened 2023-02-26
Organization proposal

leoheck:
I am proposing to organize the repo keeping python files inside a folder. 
It keeps the root path (and the initial page of the GitHub) shorter
It also helps to find python files easier. 

This also adds `anytree` as dependency so it will be listed with the Addon Manager as it was stated here:
https://forum.freecad.org/viewtopic.php?t=76394

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # FreeCAD Assembly 4 workbench
 
-Current version 0.12.7-dev
+Current version 0.12.6
 
 
 ## Overview


Zolko-123:
Too much work for no identifiable gain

Zolko-123:
> This also adds `anytree` as dependency so it will be listed with the Addon Manager

please do not mix different functionalities in 1 PR

leoheck:
It is just a single work. Organization (moving files to a new folder)... for that, some paths were updated too.

leoheck:
I will split this update in 2 then

PR #422, opened 2023-04-09
ExpressionArray, support for PartDesign::Plane

Jolbas:
Great for mirroring in datum plane
https://forum.freecad.org/viewtopic.php?p=674553


Zolko-123:
nice, thanx

PR #427, opened 2023-04-17
Fix typos and trim trailing whitespace

luzpaz:
To quickly review changes you can choose to ignore whitespace in the review (see https://cloudfour.com/thinks/quick-tip-how-to-hide-whitespace-changes-in-git-diffs/#github) to see typos.


Zolko-123:
You have been toying with some automatic format analyzer lately ? Why-not, but holy-cow, 34 files changed, and PR directly into master ? You should know better by now

PR #429, opened 2023-04-22
Add initial interference checks

leoheck:
Hi Zolko, this PR is my initial attempt to add my interference check script to this workbench.
The code should be improved of course, and also the idea can get matured by discussing it here.

In short, this adds 2 toolbar icons to Checks Interferences and also Removes the generated Interference check objects. The icons and the place on the toolbar should be also improved later, I just added them to have something to start the discussion.
image

The idea is quite simple. The interference check iterates through all objects inside the Model/Assembly and performs a `Part > Common` operation with each part against the others. The resulting object that has a volume equal to zero, does not have parts interfering and is excluded. The objects that have a volume greater than zero are interfering with each other. Then this "common" object is highlighted. The name of each common object is related to both parts that conflict, so it is easy to locate them to fix them. 

Every time the user hits the toolbar button to check the Interference folder is recreated. The operation may take some time.

The second button removes the Interference folder easily since just deleting it by hitting the Del key does not remove all objects from inside of this folder by default. It also restores the visibility of the Model/Assembly. It could also restore the initial visibility of everything but this is not implemented yet.

A lot of cleanup and improvement of the code can happen since I quickly ported my initial script to this workbench. 

The Interference folder has also a copy of the Assembly with different opacity so it is easier to locate visually the interferences where they occur. 

There are also objects missing from the check (maybe fasteners, and others) that would be good to have included.

Here is an example (a caliper with an object next to its end)
image

The interferences are highlighted here, after processing... 
image

The object tree with the objects created
image

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -150,6 +150,8 @@ class Assembly4Workbench(Workbench):
         self.dot()
         import makeBomCmd          # creates the parts list
         self.dot()
+        import checkInterference   # check interferences btween parts inside the Assembly
+        self.dot()
         import exportFiles         # creates a hierarchical tree listing of files in an assembly
         self.dot()
         import HelpCmd             # shows a basic help window
@@ -235,6 +237,8 @@ class Assembly4Workbench(Workbench):
                         "Asm4_makeLocalBOM",
                         "Asm4_makeBOM",
                         "Asm4_listLinkedFiles",
+                        "Asm4_checkInterference",
+                        "Asm4_removeInterference",
                         "Asm4_Measure",
                         'Asm4_showLcs',
                         'Asm4_hideLcs',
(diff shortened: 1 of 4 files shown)


Zolko-123:
This is great, thanx. Didn't test it yet.

I'm not sure about the second tool: I think that we should only check for interferences, it's up to the designer to do something about it or not. Think a spring or moving objects ...

As for fasteners, no that wouldn't work because of the threads.

leoheck:
For fasteners the Common operation would work, actually. If the user wants the thread or the screw wider than the hole as I usually do since it works when I 3d print things (I make holes are 10% smaller than screws), then he will see interferences and he can interpret as a good thing. Now if the screw has to pass through the holes the user is going to expect no interference. It is just how he interpret the results. The interference checks against the results could be stored in another folder to make the decision process easier. I didn't include them right now since the operation is a time consuming since each part is checked against all other parts.

Well, let me know if you think this feature is promising when you use it. Hope it works for you I the first try too.

Zolko-123:
OK, so yes it works, nice, well done.

Looking at the code, I see that it only checks for App::Link objects:

```
if obj1.Visibility == True and obj1.TypeId == 'App::Link':
```

So that won't see pure parts, and not variant links and the like. 

Also, for each object (App::Link) it creates a new Part::Feature and you set its shape to that of the target. Do you think we could use ShapeBinders instead ? I tried to do that by hand and it seems to work. Dunno if it's faster or not though

Zolko-123:
I update the repo with my modifications

PR #432, opened 2023-05-07
proof of concept modified only newAssembly command

onekk:
Here my proposal.

not more complications, only I've added a d_doc variable that could be used as comment in code, if needed or wanted.

Regards

Carlo D.


Zolko-123:
OK,  but this defeats the purpose of a library: there is only 1 command `newAssemblyCmd` that would call this function `create_assembly()`. And if someone wants to see what that command does, he needs to look at 2 files instead of 1. 

Is it possible to place that function in the same file as the command, so that everything is in 1 file ?

onekk:
Yes and No as You have to import only Asm4libs to make a decent scripts.

But Ams4 libs to be imported in many places, so IMHO is not a strange place.

Regards

Carlo D.

Zolko-123:
thank-you, but no

PR #437, opened 2023-05-08
Improvements of Interference Checks with GUI

leoheck:
Thanks for your last fix. I think we are converging on a nice tool.

This time I fixed the missing test I removed last time by mistake, that was important. 

I also improved the console messages. It will print all the comparisons (including the replicated ones) since it is easier for the user to find them and to make sure all the checks are being performed. It will also add under the comparison the status if it is clear, colliding, touching, or already processed. 

Also, if there is no interference, it now removes the folder, restores the visibility of the Model, and prints a message saying the model is clear. (maybe a popup would be better in this case?)

### DIFF
--- a/checkInterference.py
+++ b/checkInterference.py
@@ -15,6 +15,7 @@ import FreeCADGui as Gui
 import FreeCAD as App
 
 import Asm4_libs as Asm4
+import showHideLcsCmd as lcs
 import Part
 
 MIN_VOLUME_ALLOWED = 0.0001
@@ -45,6 +46,7 @@ class checkInterference:
         self.modelDoc = App.ActiveDocument
         # this has been checked before
         self.model = Asm4.getAssembly()
+        lcs.showHide(False)
         self.remove_interference_folder()
         self.check_interferences()
         self.modelDoc.recompute()
--- a/checkInterference.py
+++ b/checkInterference.py
@@ -11,6 +11,7 @@ import math
 import logging
 
 from PySide import QtGui, QtCore
+
 import FreeCADGui as Gui
 import FreeCAD as App
 
@@ -29,6 +30,7 @@ class checkInterference:
         self.current_value = 0
         self.max_value = -1
         self.progress_log = str()
+        self.enable_fasteners_check = False
 
     def GetResources(self):
         menutext = "Check Intereferences"
@@ -381,16 +383,28 @@ class checkInterference:
         # do something to abort current operations, if any
         self.abort = True
         self.progress_log = ""
-        self.log_area.setPlainText(self.progress_log)        
-        self.remove_interference_folder()        
+        self.log_area.setPlainText(self.progress_log)
+
+        # TODO: if running only
+        # self.progress_log += "\nOPERATION ABORTED\n"
+        # self.log_area.setPlainText(self.progress_log)
+        self.remove_interference_folder()
         self.model.Visibility = True
         self.modelDoc.Parts.Visibility = True
         self.UI.close()
 
     def onClear(self):
         self.progress_log = ""
-        self.log_area.setPlainText(self.progress_log)        
+        self.log_area.setPlainText(self.progress_log)
         self.remove_interference_folder()
+
+
+    def onCheckFasterners(self, state):
+        if state == QtCore.Qt.Checked:
+            self.enable_fasteners_check = True
+        else:
+            self.enable_fasteners_check = False
+
 
     def progress_bar_reset(self):
         self.current_value = 0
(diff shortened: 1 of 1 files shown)


Zolko-123:
The problem with `print()` is that it doesn't print to the _Report View_, so the user doesn't see the messages. Ideally, we should open a window with a progress bar. Do you think we could estimate the progress of the calculation ?

leoheck:
Hm, did not understand. For me, it does print on Report View.

image

leoheck:
I was about to show you using a bigger project and found and issue. I will try to fix it.

leoheck:
I am going to try to add this percent of comparisons to this PR so we can discuss later, cool?

leoheck:
> The problem with print() is that it doesn't print to the Report View,

Ah, I see what you are saying. You are saying that it does not print while processing. And if you run freecad from the terminal we can see it being printed on each interaction. Got it.

leoheck:
Sometimes when there is no visible part inside a Part, its copy does not produce a visible shape, then the Common is not computed. I improve that. 

image

Now I will try to work on a progress bar/popup.

Zolko-123:
> Ah, I wanted to turn this block here into a function. Since it is used twice. But for some reason, I could not make it work.

Did you look at my version of the file ? 

https://github.com/Zolko-123/FreeCAD_Assembly4/blob/development/checkInterference_z.py

I solved the problem differently: first all shapes are copied (line 72 and `make_shape_copy()` ), then all interferences are checked (line 84 and `parse_interferences()` ) and finally the results are analysed as before (line 88)

Since the number of parts is know – N – the number of interference checks is also known – N*(N-1)/2 – so we should be able to show a progress bar based on that

leoheck:
And we could add an abort button too, =)

leoheck:
I did not look your file, I am going to check it later, today.

leoheck:
Ah, found  that this plane, is generated, sometimes, when the LCS is visible. 
How can I hide the LCSs on python?
image

leoheck:
I am starting to have something.
image

leoheck:
@Zolko-123 I was playing around with the GUI and the progress bar. 
It is not ready to merge since I still have to clean the code and improve the info.
But it is getting interesting.

leoheck:
Just a quick idea how it looks like. Tomorrow I will improve it more. 
image

Zolko-123:
Nice. The "check fasteners" buttons is good idea, although it must not produce an error if the fasteners workbench is not installed. You should leave it for a later improvement. as for the lower buttons, I suggest only 1: "Abort" which becomes "Close" when the calculation is finished. The interferences check is launched right at the beginning, no need for an extra button

Also, I think that [my version](https://github.com/Zolko-123/FreeCAD_Assembly4/blob/development/checkInterference_z.py#L130) of `parse_interferences()` is much shorter and clearer than [the original](https://github.com/leoheck/FreeCAD_Assembly4/blob/interference-check-improvements-NEW/checkInterference.py#L62) `check_interferences()`. In my version, it is done in 2 steps: 1) the shapes of all objects of the correct _TypeId_ (PARTID2CHECK) are duplicated in the _ShapesCopy_ folder, and 2) then these copies are parsed 2-by-2 and the interferences created. I find the data flow of the original function quite complex

leoheck:
Yeah, Insaw your version. I am trying to organize and improve this popup, and later I can improve it to have a.better code. 

Considering fasteners, it doesn't matter if the workbench is installed. it is converted to a Part to be compared. The file cab have fasteners even if the user does not have fasteners workbench installed. but if I am wrong, we could disable the fasteners checkbox of the workbench is not present

leoheck:
I am playing learning and improving the GUI of this feature, did not touch the interference checks yet.

I am committing the code to save the progress. I invite you to check it out to see the progress and discuss the GUI.

I am looking for a way to 
- add colors in the Log View
- scroll the log view to the bottom automatically.

Do you know to do any of those?

image

leoheck:
Alright, I could make it work better. It is more mature now. 

Now it is nice to give it a try to see if it is working for you too, @Zolko-123 
And maybe, to check if it is working with different Assemblies that have more and different types of objects.

Maybe it is good enough to ask people to try it on Freecad Forum, too.

Some things I still want to improve (mostly the GUI).
1. TextEdit has to scroll down automatically with new text.
1. Text can be copied but not edited, I should disable that somehow. 
1. Colors on the `log_view`, would be nice to have. 
1. Set the settings box with a fixed height. So it would work fine when resized.
1. Buttons could be smaller and have a fixed size. 
1. The Abort button could have a red image or something.

A sample of the current thing, here the operation was aborted.
image

Here it ended nice
image

leoheck:
Hi @Zolko-123 , I believe the Interference Check is working quite well now. 
It it is working ok for you without any other request, we could merge this code, to check how it will work for other users too.

Zolko-123:
OK will do

Zolko-123:
Thanks. This is a big chunk so I'll review it and make a new release with that. Then we can go for the BOM

leoheck:
Sure, that would be awesome. I would like to keep changes small, unfortunately, I wanted to improve the whole process this time.

PR #440, opened 2023-05-27
[WIP] BOM improvements with GUI/ProgressBar and User Control

leoheck:
This is a work in progress still, not ready to merge it.

### DIFF
--- a/InitGui.py
+++ b/InitGui.py
@@ -232,7 +232,6 @@ class Assembly4Workbench(Workbench):
                         "Asm4_shapeBinder",
                         "Separator",
                         "Asm4_infoPart",
-                        "Asm4_makeLocalBOM",
                         "Asm4_makeBOM",
                         "Asm4_listLinkedFiles",
                         "Asm4_checkInterference",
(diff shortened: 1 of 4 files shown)


Zolko-123:
thank-you, but I'd like to split this into 2 merges: 1 for PartInfo, and 1 for makeBom. So ideally we would make a PartInfo function where the user can set/see/retrieve the relevant informations to the part, may-be using templates, but independently from any BOM (and leave BOM broken for the while, may-be even removing the button temporarily), and when that is done **THEN** go for the BOM. 

This BOM has been half-way working for so long that I'd like to take the time to do it well this time.

What do you think: can we split this first for PartInfo ?

leoheck:
Sure, that would be perfect. I had to make changes in the PartInfo since I wanted to fix the whole functionality. But having the PartInfo well defined and coded would help me focus the makeBom easier.

Once the PartInfo is done and merged, I can adjust the makeBom to use it.

Zolko-123:
Hello, 

I'm sorry but I can't manage such huge PRs, I don't have the time or the competence. I'm closing this but please propose smaller PR they are welcome

PR #446, opened 2023-07-15
Low hanging fruit bug fixes

JonasThomas:
I think this might be my first pull request... 
Let me know if you need additional info.

### DIFF
--- a/infoKeys.py
+++ b/infoKeys.py
@@ -3,7 +3,6 @@
 #
 # LGPL
 
-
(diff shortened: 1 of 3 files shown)


Zolko-123:
Thank-you for your help, but as it is this PR included too many changes. Please separate really trivial fixes from more complicated functional code

JonasThomas:
I was working on this when I was between jobs.  At this point, my new position is sucking up my band width.

JonasThomas:
> @JonasThomas Any news on this? I think that those bug fixes would be great to get included - Thank you for your work! :)
@Tectu
I was working on this when I was between jobs.  At this point, my new position is sucking up my band width and haven't had time to look at it.

Zolko-123:
bah, since this only changes BoM and PartInfo which are broken anyway, why not try this one

PR #448, opened 2023-08-10
Attempt to prevent bad variable names

madhead:


Zolko-123:
Why not, it's not a ad idea after all. Next time please make PRs against the development branch

Zolko-123:
The test was inverted, "Angle" gave an error while "A" was OK, "time" gave an error while "t" was OK. Fixed in the latest 0.50.4 release, can you please check ?

madhead:
Ok, I have this version installed

image

And it works as expected, thanks!

image
image
image

PR #460, opened 2023-12-27
Fix state serialization on FreeCAD >= 0.22

andir:
As of https://github.com/FreeCAD/FreeCAD/commit/83d4080fe8 FreeCAD (for compatibility with Python 3.11) has renamed the methods to loads and dumps. Using this workbench with the latest FreeCAD main branch surfaced this issue as I wasn't able to save my documents again.

I found the change https://github.com/fra589/CurvesWB/commit/3374dbb83ea2a09787819b1e153a5697f5e19ea7 in the CurvesWB and adopted it for this WB.

### DIFF
--- a/Asm4_objects.py
+++ b/Asm4_objects.py
@@ -20,6 +20,8 @@ from FreeCAD import Console as FCC
 import Asm4_libs as Asm4
 
 
+FREECAD_0_22 = (int(App.Version()[0]), int(App.Version()[1])) >= (0, 22)
+
 
 """
     +-----------------------------------------------+
@@ -51,12 +53,20 @@ class VariantLink( object ):
     def __init__(self):
         FCC.PrintMessage('Initialising ...\n')
         self.Object = None
-    
-    def __getstate__(self):
-        return
 
-    def __setstate__(self,_state):
-        return
+
+    if FREECAD_0_22:
+        def loads(self, state):
+            return None
+
+        def dumps(self):
+            return None
+    else: 
+        def __getstate__(self):
+            return None
+
+        def __setstate__(self, _state):
+            return None
     
     # new Python API for overriding C++ view provider of the binding object
     def getViewProviderName(self,_obj):
@@ -237,11 +247,18 @@ class ViewProviderVariant(object):
             iconPath = os.path.join( Asm4.iconPath, 'Variant_Link.svg' )
         return iconPath
                 
-    def __getstate__(self):
-        return None
+    if FREECAD_0_22:
+        def loads(self, state):
+            return None
 
-    def __setstate__(self, _state):
-        return None
+        def dumps(self):
+            return None
+    else: 
+        def __getstate__(self):
+            return None
+
+        def __setstate__(self, _state):
+            return None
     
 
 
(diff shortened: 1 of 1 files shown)


Zolko-123:
thank-you for the contribution, but I don't understand what it does. Can you please explain ?

andir:
> Since Py3.11 the methods names __setstate__ and __getstate__ conflict with the method names added to the object class.
Thus rename them to 'loads' and 'dumps'

(from the FreeCAD commit mesage).

This means that the old `__setstate__` and `__getstate__` methods aren't working with newer version of FreeCAD anymore. FreeCAD decided to rename them to `dumps` and `loads` instead. What my changes are doing is, depending on the FreeCAD version the WB is running setting either the new or the old functions.

Zolko-123:
> FreeCAD decided to rename them to dumps and loads instead

no, that was not a FreeCAD decision, that's a decision by the `pickle` Python module. That [commit and code](https://github.com/FreeCAD/FreeCAD/commit/83d4080fe8) are so wrong on any scale that I cannot accept that. 

Also, when looking at the `pickle` documentation:

> _Warning : The pickle module is not secure. Only `unpickle` data you trust. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling_

Also, v0.22 is still in development, hopefully Werner will wake-up and come with a serious solution instead of this half-baked incompatible horror show

Zolko-123:
according to PR https://github.com/FreeCAD/FreeCAD/pull/10581 this is not necessary :

> App: still support __getstate__/__setstate__ for add-ons for < Py3.11 

`// support add-ons that use the old method names`

Syres916:
@Zolko-123 it **_is_** necessary for Python 3.11 and greater as _setstate_ is now a reserved attribute brought in by the Python developers. I'm trying in my spare time to get all the Addon authors to have this fix implemented asap because of the number of Linux distros having Python 3.11 as default and no doubt 3.12 is not far from being the new standard. It's up to you but all you're going to do is get more complaints from users trying to save their files and receiving a sea of red in the Report View.

Syres916:
@andir just some feedback on your implementation, there is no need to have the FreeCAD version check especially as version 0.21.2 has the _dumps/loads_ in it and 0.21.3 will have the complete json code fix. You can just duplicate the code as per https://github.com/jaheyns/CfdOF/commit/6be2891a38ed67969216f7e7b6449f7c9684a8c9

Zolko-123:
> I'm trying in my spare time ...

you think I'm paid for the time I spend on Assembly4 ?

Zolko-123:
1) This problem doesn't depend on the FreeCAD version but on the Python version. This PR doesn't solve the problem

2) **realthunder** [proposes a solution](https://github.com/FreeCAD/FreeCAD/commit/83d4080fe86a8d98dfab87763623f8a1bd76bf43#commitcomment-136214612) in FreeCAD core that solves the problem where it should be solved : in the core, making FreeCAD backward compatible without this ugly hack. 

See also https://github.com/FreeCAD/FreeCAD/issues/10460 . Therefore I must refuse.

Issue #254, opened 2021-10-01
Translation


FarmingSoul:
How make a update of \Resources\translations\asm4.ts for continue to translate the new dialog ??
Thankx all

Zolko-123:
I don't know

Zolko-123:
translation was added by @Thowil :

https://github.com/Zolko-123/FreeCAD_Assembly4/pull/214/commits/5005f483fa6e103d4f47fc98d2bcce1503b89b8f

I don't know how it works

hasecilu:
Hello, I can implement the translation for the WB, could you reopen the ticket?
I have a minimal implementation working.

Zolko-123:
done

Zolko-123:
hello @hasecilu , I've merged your translation code and it works indeed, thank-you for that :

> Inicializando en entorno de trabajo Assembly4 (0.50.8-dev) ............................. hecho.

I've modified the code to better suit my personal tastes (the function is called Qtranslate as before), so please update before making further changes. But I have 2 issues:

1) on my system, `pylupdate6` doesn't exist, so I can't update the language pack

2) more profound: I don't think that changing the menu entries is really that helpful, translating the explanatory tooltip would be much better for the user:


![translation](https://github.com/Zolko-123/FreeCAD_Assembly4/assets/46596542/5076ff79-a18a-4a3d-b9c3-52824c61c52c)


Changing the menu entry brings the disadvantage that the user cannot make an Internet search easily, and the default English tooltip is not translated. What I'd suggest is the opposite: leave the menu entry the default and provide translations for the tooltips. Is this doable ?

hasecilu:
> I've modified the code to better suit my personal tastes (the function is called Qtranslate as before), so please update before

As I mentioned on the PR review is not possible to use Qtranslate() because pylupdate is looking for instances of `translate(context, string)`, I tried with pyside2-lupdate, pylupdate5 and pylupdate6 and none of them catch Qtranslate().



> 1. on my system, `pylupdate6` doesn't exist, so I can't update the language pack

You can use `pylupdate5`, I used 6 because I found a problem with version 5 that doesn't pick strings: https://github.com/pyqt/python-qt5/issues/54
But if here we're not formatting the code probably will not be any problem.



> 2\. I don't think that changing the menu entries is really that helpful, translating the explanatory tooltip would be much better for the user:

[...]

PR #464, opened 2024-02-06
WIP: Add translation support, initial testing

hasecilu:
The previous attempt to support translation was not working. Now a proven methodology is implemented. (This method is tested on Fasteners, SheetMetal, Gear (open PR), FreeGrid, starterkit)

From initial testing is working as expected.

I think I can continue the implementation next week or in 2 weeks. Meanwhile is someone is interested on finishing it just need to:

```python
from TranslateUtils import translate

# Replace strings with translate(context, string)
            self.menutext = translate("Fasteners", "Insert Screw")
```

Translator need to execute bash script to update available strings. Please read script. `./update_translation es-ES`

image

Related #254


Zolko-123:
thank-you for this contribution, BUT:

1) make PRs against the development branch please, I never accept PRs directly into master

2) make a smaller PR for an initial implementation as I have to review all the files. I suggest you only do _Fasteners_ and _newPart_ for example. If that works well we can go for a larger PR

hasecilu:
Hi, somehow I mis-readed the 2nd point and ended marking more strings, I marked the title of the next tools

image

For marking multiline strings for translation, since we want the whole paragrpah to be together we need to do it like this:
```python
```
Trying to use `\` to split the strings or using `+` to concatenate the strings doesn't work.

Have you thought about using some kind of code formatting tool? Some changes here got affected and move some character that seems to be manually formatted...

Zolko-123:
I don't see any obvious reason why the file Asm4_libs should be modified for translation, and the code doesn't seem to include any such things. Why did you include it with this PR ?

If it's code formatting and white-space cleanups, why not, but that should be part of another PR

Zolko-123:
> from DraftTools import translate

this no in no way because it imports all of Draft

Asm4_Translate.py
Zolko-123, line 1: Please don't delete this file yet

FastenersLib.py
Zolko-123, line 1: OK

InitGui.py
Zolko-123, line 1: Please don't modify this file with this PR

TranslateUtils.py
Zolko-123, line 1: I'd rather prefer that the function be called Qtranslate as today:

`Qtranslate = FreeCAD.Qt.translate`

This way there are less changes in the rest of the code

gotoDocumentCmd.py
Zolko-123, line 1: from TranslateUtils import Qtranslate

infoPartCmd.py
Zolko-123, line 1: Please remove this file change from this PR

newAssemblyCmd.py
Zolko-123, line 1: from TranslateUtils import Qtranslate

newDatumCmd.py
Zolko-123, line 1: Please remove this file change from this PR

variantLinkCmd.py
Zolko-123, line 1: Please remove this file change from this PR

hasecilu:
Hi, I pushed most of the requested changes, above is the reason why some requested changes were not made.

> > from DraftTools import translate
> 
> this no in no way because it imports all of Draft.
> 
> **EDIT**: I see that it's in an `if...else` loop : when would this condition be False (and trigger the DraftTools translate) ?

It doesn't import all moduloele, only `translate`.

I think would go to the `else` with old FreeCAD version, maybe very old.

The code is taken from BIM WB by Yorik: https://github.com/yorikvanhavre/BIM_Workbench/blob/master/BimTranslateUtils.py

PR #465, opened 2024-02-09
Make license match SPDX ID

chennes:
See https://spdx.org/licenses/ -- note that you might actually mean GPL-2.0-or-later, in which case use that instead.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -5,7 +5,7 @@
   <version>0.50.7</version>
   <date>2024.01.31</date>
   <maintainer email="zolko@zzaero.com">Zolko-123</maintainer>
-  <license file="LICENSE">GPL-2</license>
+  <license file="LICENSE">GPL-2.0-only</license>
   <url type="repository" branch="master">https://github.com/Zolko-123/FreeCAD_Assembly4</url>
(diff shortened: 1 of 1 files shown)


Zolko-123:
is this important ?

chennes:
The addon manager now has additional support for filtering on licenses: unrecognized license strings are treated as unlicensed (e.g. All Rights Reserved), and users may choose to not show those addons. In this particular case I've written some heuristics that will probably recognize it, so this is a low-priority fix.

Zolko-123:
> unrecognized license strings

is there a list of **recognized** license strings that one can consult to choose from ?

chennes:
It's the "identifier" column here: https://spdx.org/licenses/

Zolko-123:
> The addon manager now has additional support for filtering on licenses

So this is a self-inflicted pain ? Looks furiously like a solution in search of a problem

3x380V:
_The purpose of the SPDX License List is to enable efficient and reliable identification of such licenses and exceptions in an SPDX document, in source files or elsewhere._ 

Care to reconsider?

PR #469, opened 2024-03-04
save/restore mirrored objects in configurations

zisoft:
While working on my current project I noticed that ASM4 mirrored objects are not properly saved/restored in configurations.


Zolko-123:
Good catch, thank-you

PR #477, opened 2024-03-19
v0.50.10

Zolko-123:


makeArrayCmd.py
```diff
@@ -209,6 +209,8 @@ def _setupProperties(self, obj):
         obj.setExpression('.Placer.Rotation.Angle', '180 * (Index % 2)')
         obj.setPropertyStatus('Placer', 'Hidden')
         obj.setPropertyStatus('Scaler', 'Hidden')
+        # https://github.com/Zolko-123/FreeCAD_Assembly4/issues/474
+        Asm4.makeAsmProperties(obj) ?
```
marc0s, line 213: Been a while since a did any Python, but I think this `?` shouldn't be there.

```
17:52:03  Initializing Assembly4 workbench (0.50.10) .............invalid syntax (makeArrayCmd.py, line 213)
```

I've *just* installed Assembly4 and couldn't try it because of this invaild syntax error

PR #479, opened 2024-03-19
Fix typo

hasecilu:
I was doing some rebase on translation when this broke the update.


Zolko-123:
yes, thank-you, I fixed it in a hurry

PR #516, opened 2024-09-15
Fix interference checks

leoheck:
This feature was not working after the removal of the Parts folder.


leoheck:
Fixing this feature allowed me to fix my design before sending it to the production last night. This is a really helpful feature. More people should know about it =)

leoheck:
The latest commits improved the interference draw style to show the line, so it will be easier to find places where the collision is quite small which is actually the goal of the tool since when the colision is too obvious the user will catch easily without the tool.
image

Zolko-123:
Thanks

on another note, I have seen that your algorithm to parse all the parts was not optimal, and implemented something faster, that was in the __zh_ file. Did you have a look at it ?

leoheck:
No, sorry. Didn't know that.

leoheck:
Improving the algorithm will be a nice thing. I will try to understand your code.

Also, I was thinking that if I could avoid the common operation by comparing if the bounding boxes overlap... this would speed up things.

PR #530, opened 2025-02-08
Creates LCS inside a body if features of the body are selected

leoheck:
This is outaded.

~Hi, @Zolko-123 this PR improves the Create Datum process.~

~After creating a new Datum, the selection (selected object) is changed, making the process of creating multiple datums a pain.~

~With this fix, it is possible to create multiple Datums on the same object in series since it preserves the original selection.~

~Here is a video of the process after the fix where the selection returns to the Body after each Datum is created.~

https://github.com/user-attachments/assets/35fa55eb-94fd-40e1-adba-4aeb62055baf

### DIFF
--- a/newDatumCmd.py
+++ b/newDatumCmd.py
@@ -87,14 +87,11 @@ class newDatum:
             selectedObj = Gui.Selection.getSelection()[0]
             # ... and it's an App::Part or an datum object
             selType = selectedObj.TypeId
-
             if selType in self.containers or selType in Asm4.datumTypes or selType=='Sketcher::SketchObject':
                 return(selectedObj)
-
             # When the selected item is a feature of a Body, it returns the Body itself.
-            if selectedObj._Body.TypeId == 'PartDesign::Body':
+            elif obj.getParentGeoFeatureGroup().TypeId in Asm4.containerTypes:
                 return(selectedObj._Body)
-
         # or if nothing is selected ...
         elif Asm4.getAssembly():
             # ... but there is as assembly:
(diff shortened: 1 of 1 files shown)


Zolko-123:
Interesting ... I'll merge the translation PR and come back to this one

leoheck:
I have a better solution for this. The best solution is to check if the feature is inside a body and if it is so, then it uses the Body as the selected item.

Zolko-123:
The thing with a timer looked to be a source of bugs, this seems better

Zolko-123:
But will that work with Part containers ? Does that "_Body" attribute always work ? It doesn't with 0.21.3, is it a new feature ? Then it will need something else, like `obj.getParentGeoFeatureGroup().TypeId` . Something like : 

`if obj.getParentGeoFeatureGroup().TypeId in Asm4.containerTypes` 

It's the same but more general, will work on all containers

Zolko-123:
I like the idea, can you please update your code ?

leoheck:
Sure, I will update later, ok?

leoheck:
Thanks for the suggestion @Zolko-123 this seems better indeed.

Zolko-123:
I think you have mixed things up. Shouldn't this rather be:

```
            elif selectedObj.getParentGeoFeatureGroup().TypeId in Asm4.containerTypes:
                return(selectedObj.getParentGeoFeatureGroup())
```
?

leoheck:
yeah, I was about to test it. it is working on 0.21.2, tested now.

leoheck:
it also works with 1.0

leoheck:
and 1.1

Zolko-123:
thanx

PR #532, opened 2025-02-14
Click on 3d model to edit placement

leoheck:
Hi @Zolko-123 

Similar to my last PR where I improved the LCS creation sequentially, here I made it possible to edit the placement by clicking the 3d model instead of selecting the model in the Object tree.

I tested this on `0.21` and `1.1-dev`

I am attaching this sample file so you can easily verify if it works. 
[immediate_parent_link_selection.FCStd.zip](https://github.com/user-attachments/files/18791504/immediate_parent_link_selection.FCStd.zip)

The test is simple, click on each object, and check if it enables the `Asm4_placeLink` icon and if its function works as expected.

There are 5 different objects only Links_1, Link_2, and Link_3 should enable the feature.
image

This is my first attempt after working on this for some time, so things can be improved to make it look more like the Asm4.

### DIFF
--- a/Asm4_libs.py
+++ b/Asm4_libs.py
@@ -682,6 +682,19 @@ def getSelectedContainer():
     return retval
 
 
+def getDirectParent():
+    # including links, specially usefull for when selecting ojects by clicking on 3d models
+    sels = Gui.Selection.getSelectionEx("", 0)
+    sel = sels[0]
+    sub = sel.SubElementNames[0] if sel.SubElementNames else ""
+    obj = sel.Object.getSubObject(sub, 1)
+    if obj == sel.Object:
+        direct_parent = None
+    else:
+        direct_parent = sel.Object.getSubObject(sub.rpartition(obj.Name)[0], 1)
+    return direct_parent
+
+
 # returns the selected App::Link
 def getSelectedLink():
     retval = None
--- a/placeLinkCmd.py
+++ b/placeLinkCmd.py
@@ -39,16 +39,26 @@ class placeLinkCmd():
                 }
 
     def IsActive(self):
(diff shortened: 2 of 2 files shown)


leoheck:
This behavior is a bit annoying indeed.

I put this issue here https://github.com/FreeCAD/FreeCAD/issues/19564 and Roy_43 gave me the solution in the forum to find the direct parent of a selected feature that includes links https://forum.freecad.org/viewtopic.php?t=94767

As far as I could test it is working. But I will be using it this week-end to make sure it works on a real project.

And if this works, then it may be used in other places since I just added to the place link, there may be other places that could use this.

leoheck:
There was something strange I was using the `placePartUI` instead of `placeLinkUI`. Now it looks better. I could also improve the `IsActive` process.

leoheck:
What do you think about this improvement @Zolko-123 .
I have been using it for 3 weeks already, it is working smoothly.

leoheck:
This PR works for Bodies, but not yet for Parts.

Zolko-123:
Did you try to use the selection filter ? I think it does what you want :

selfilter

leoheck:
oh, did not know about this feature, I will have a look

leoheck:
@Zolko-123 now it is working nicely for Bodies and Parts. Tested with `0.21` and `1.1-dev`.

Zolko-123:
Ideally, I'd like this to work with the selection filter : if the "Part" is toggled, then the selection selects the parent Body/Part/Link even when a particular feature/geometry is selected in the 3D window, else it behaves as the normal selection

leoheck:
This branch was updated with the latest development, and it is working.

PR #536, opened 2025-02-16
Improve PlaceLink with Equations for Translation, and custom step for Rotation

leoheck:
Preview of the options.
This PR should have been added some time ago, now I took the opportunity to push this. 

Translations can be performed with equations.
Rotations can be customized to have a custom step.
And can go in both directions, it is also possible to rest the rotation on each axis.
By default, it is set to steps of 45, instead of the current 90.

image

Here in this file `insertLinkCmd.py` I am enabling double click on object names to place a link.

### DIFF
--- a/Asm4_libs.py
+++ b/Asm4_libs.py
@@ -748,4 +748,106 @@ class QUnitSpinBox(QtGui.QDoubleSpinBox):
         return super().setValue(
             distance / self.length_divisor,
         ) 
-        
+
+
+#=========================
+
+# https://github.com/FreeCAD/FreeCAD/blob/af1f7385d47b9887e758794afdc3b1f69b406b96/src/Mod/CAM/Path/Base/Gui/Util.py#L114
+# https://github.com/gbroques/freecad-rotations
+
+from math import cos, radians, sin, acos, degrees, sqrt
+from typing import Tuple
+
+def euler_to_quaternion(yaw: float,
+                        pitch: float,
+                        roll: float) -> Tuple[float, float, float, float]:
+    """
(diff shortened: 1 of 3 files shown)


Zolko-123:
the free rotation angle has been proposed many times but it doesn't work, because you can't combine them. It has to be 90° only. If you need any random rotation angle you should insert an LCS (attach that LCS and make it rotate around the axis you need, and attach your part to that LCS).

So I can't take that.

But the expression in the translation widget is good, can you please make a separate PR for it ?

Thank-you for the proposal but I have to close it

leoheck:
It is not free rotation angle. It is a customized step for rotation, the rotation will happen in custom steps. The PR works just fine, did you have time to check? I can make a video to show it working.

leoheck:
Please, take a look.

Here I have 3 links, where I rotate each one using 90, 15, and 1 degree.

Each one was translated using equations on the X-Axis.
`0, 1*80, 2*80`

Uploading asm4_link_placement-2025-02-16_21.01.25.mp4…

Zolko-123:
yes I know it works the first time you do it, but what happens when you edit the placement of the second part again ? And when you attach it to or with some other LCS ?

We have done this before and it didn't work then. See https://github.com/Zolko-123/FreeCAD_Assembly4/pull/325

Also, what would be the purpose (even if we managed to get it working) ? Please remember that Asm4's principle is KISS (Keep It Simple and Stable) , the very opposite of what FreeCAD does currently. The integrated assembly workbench will be long forgotten when Asm4 will still be working thanks to this. We're not running after features, we're aiming for (very) long term stability. But I thank-you for all your code and suggestions, they make this healthy

PR #537, opened 2025-02-16
Add possibility of shortcuts.

leoheck:
This PR adds the possibility of shortcuts to some useful functions. 
For some, there is a suggestion of a shortcut.

If the `Accel = ""` then there is no shortcut, but it will be able to add it by the user.

### DIFF
--- a/insertLinkCmd.py
+++ b/insertLinkCmd.py
@@ -44,6 +44,7 @@ class insertLink():
         tooltip += "Select the broken link, launch this command, and select a new target part in the list</p>"
         iconFile = 'Link_Part.svg'
         return {"MenuText" : "Insert Part", 
+                "Accel"    : "i",
                 "ToolTip"  : tooltip, 
                 "Pixmap"   : os.path.join( Asm4.iconPath , iconFile )
                 }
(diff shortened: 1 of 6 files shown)


leoheck:
So with this and the other PR is possible to click on a 3d model and hit e to edit the placement.

Zolko-123:
good idea. Are you sure it doesn't interfere with shortcuts from other workbenches, or main FreeCAD ? Or is it limited to the workbench in use ?

leoheck:
Thanks.

I am adding a single-key shortcut.
I haven't seen those being used, freecad uses mostly 2 key shortcuts for the core functions.

I added shortcuts for the following items.

1. New Assembly (A), the same used in the new standard Assembly Wb. So this one is ok.
2. Create Part (P), Body (B), Group (G), 
3. Insert (I) Link
4. New LCS, Point, Axis, Plane, and Sketch, do not have one, but they can be added by the user.
5. New Hole <= user can add.
6. Place Link (E) <= as in (E)dit Link
7. Show/Hide LCS <= user can add.

If you check this list, you will see the basic functionality works with 2 keys or more.
https://wiki.freecad.org/Sandbox:Keyboard_Shortcuts#:~:text=By%20arranging%20your%20workbench%20list,the%20Workbench%20Dropdown%20List%20toolbar.

PR #542, opened 2025-03-01
ThSelect the first part on the list of parts

leoheck:
This improves the usability of the insert link feature.
It automatically selects the first element of the list, especially after updating the filter.


Zolko-123:
thanx

PR #561, opened 2025-04-29
Update insertLinkCmd.py

proximalf:
Qt Parent missed. Only a problem when using a tiling window manager.

### DIFF
--- a/insertLinkCmd.py
+++ b/insertLinkCmd.py
@@ -66,7 +66,7 @@ class insertLink():
     """
     def Activated(self):
         # This function is executed when the command is activated
-        self.UI = QtGui.QDialog()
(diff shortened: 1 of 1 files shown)


Zolko-123:
nice. Should this be added to all `QtGui.QDialog` windows ?

proximalf:
This was the only instance that was bothering me, if it's a permanent window then maybe best to not have parent inheritance

proximalf:
You are correct, my bad. It works on my machine, but if I switch a non-tillable wm it doesn't work.

Zolko-123:
OK, we revert, I should have tested this more. see https://github.com/Zolko-123/FreeCAD_Assembly4/pull/567

